PR #11, opened 2024-07-18
✨ [CLI + Server] Sequin Observe w/ Websockets

RTLS:


RTLS:
* **#11** <a href="https://app.graphite.dev/github/pr/sequinstream/sequin/11?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

server/lib/sequin_web/channels/observe_channel.ex
```diff
@@ -0,0 +1,46 @@
...
+  end
+
+  def broadcast("messages:upserted", {stream_id, messages}) do
+    if any_users_present?() do
+      Logger.warning("Broadcasting messages:upserted for observation: #{stream_id}")
+      subjects = Enum.map(messages, & &1.subject)
+      messages = Streams.list_messages_for_stream(stream_id, %{subjects: subjects})
+      SequinWeb.Endpoint.broadcast("observe", "messages:upserted", %{messages: messages})
+    end
+  end
```
RTLS, line 34: we only want to load message data if there is someone connected to the websocket, else it's unnecessary load in the database

cli/cli/observe_command.go
```diff
@@ -418,9 +410,112 @@ func (s state) renderBottomBar(width int) string {
...
+	statePtr.registerCallbacks(observeChannel)
+
+	p := tea.NewProgram(statePtr, tea.WithAltScreen()) // Pass the pointer to the program
 	if _, err := p.Run(); err != nil {
 		return fmt.Errorf("error running program: %w", err)
 	}
 	return nil
 }
+
+func (s *state) registerCallbacks(observeChannel *api.ObserveChannel) {
```
acco, line 435: very cool!

acco (approved):
Per discussion yesterday, didn't look too deeply, but looks good! I like how simple the backend changes ended up being

It looks like this won't add a "tax" to CRUD on objects (i.e. "need to add controller endpoints for regular CLI commands, then some other path for observe commands") which is good - that's the biggest thing we want to avoid at this stage IMO

PR #200, opened 2024-08-31
PostHog Identify calls

thisisgoldman:
### TL;DR

Implemented PostHog user identification and improved logout functionality.

### What changed?

- Added a new `PostHogIdentify` hook in `app.js` to identify users with PostHog.
- Updated the `ConsumersLive.Index` module to push user data for PostHog identification.
- Modified the logout process in `Sidenav.svelte` to reset PostHog before redirecting.
- Integrated PostHog identification in the consumers index template.

### How to test?

1. Log in to the application.
2. Navigate to the consumers index page.
3. Check PostHog dashboard to confirm user identification.
4. Log out and verify that PostHog data is reset.

### Why make this change?

This change enhances user tracking and analytics capabilities by properly identifying users in PostHog. It also ensures that user data is cleared from PostHog upon logout, improving privacy and data accuracy.

---

### DIFF
--- a/assets/js/app.js
+++ b/assets/js/app.js
@@ -31,11 +31,26 @@ posthog.init("phc_TZn6p4BG38FxUXrH8IvmG39TEHvqdO2kXGoqrSwN8IY", {
   ui_host: "https://us.posthog.com",
 });
 
+let Hooks = {};
+
+Hooks.PostHogIdentify = {
+  mounted() {
+    const userId = this.el.dataset.userId;
+    const userEmail = this.el.dataset.userEmail;
+    const userName = this.el.dataset.userName;
+
+    posthog.identify(userId, {
+      email: userEmail,
+      name: userName,
+    });
+  },
+};
+
 let csrfToken = document
   .querySelector("meta[name='csrf-token']")
   .getAttribute("content");
 let liveSocket = new LiveSocket("/live", Socket, {
-  hooks: getHooks(Components),
+  hooks: { ...getHooks(Components), ...Hooks },
   params: { _csrf_token: csrfToken },
 });
 
--- a/assets/svelte/components/Sidenav.svelte
+++ b/assets/svelte/components/Sidenav.svelte
@@ -13,6 +13,7 @@
     ChevronLeft,
     ChevronRight,
   } from "lucide-svelte";
+  import posthog from "posthog-js";
 
   export let currentPath: string;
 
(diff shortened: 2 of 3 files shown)


thisisgoldman:
* **#200** <a href="https://app.graphite.dev/github/pr/sequinstream/sequin/200?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

 Join @thisisgoldman and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

assets/svelte/components/Sidenav.svelte
```diff
@@ -31,6 +32,11 @@
   function toggleCollapse() {
     $isNavCollapsed = !$isNavCollapsed;
   }
+
+  function handleLogout() {
+    posthog.reset();
+    window.location.href = "/logout";
+  }
```
acco, line 39: I'd prefer to do all logout-related activities in `UserLogoutLive` (mounted on `logout`) then spread out the logic here and there and elsewhere. We want to keep such logic consolidated. And it means we can redirect to `/logout` from anywhere to get this functionality.

Moving now

assets/js/app.js
```diff
@@ -31,11 +31,26 @@ posthog.init("phc_TZn6p4BG38FxUXrH8IvmG39TEHvqdO2kXGoqrSwN8IY", {
...
+    const userId = this.el.dataset.userId;
+    const userEmail = this.el.dataset.userEmail;
+    const userName = this.el.dataset.userName;
+
+    posthog.identify(userId, {
+      email: userEmail,
+      name: userName,
+    });
+  },
+};
```
acco, line 47: Hooks are the wrong pattern here. These are for when we want to attach behavior to DOM functionality. But in this instance, we just want to attach to specific events. `window.addEventListener` is a better choice

lib/sequin_web/live/consumers/index.ex
```diff
@@ -10,7 +10,8 @@ defmodule SequinWeb.ConsumersLive.Index do
   alias SequinWeb.ConsumersLive.Form
 
   @impl Phoenix.LiveView
-  def mount(_params, _session, socket) do
+  def mount(_params, session, socket) do
+    user = Sequin.Accounts.get_user_by_session_token(session["user_token"])
```
acco, line 14: We should never call this from inside LiveViews. We have helpers, namely `current_user`, for that.

acco (commented):
@thisisgoldman closing in favor of 1721495cd2fa6dbedcdaa279fa1f7040808f0eef

PR #226, opened 2024-09-06
Add Loops adapter, mailer

acco:


acco:
* **#226** <a href="https://app.graphite.dev/github/pr/sequinstream/sequin/226?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

config/runtime.exs
```diff
@@ -156,8 +158,6 @@ if config_env() == :prod do
 
   config :sequin, :dns_cluster_query, System.get_env("DNS_CLUSTER_QUERY")
 
-  config :swoosh, Sequin.Mailer, api_key: get_env.("SENDGRID_API_KEY")
```
acco, line 159: this was wrong before (should be config :sequin)

config/runtime.exs
```diff
@@ -137,6 +137,8 @@ if config_env() == :prod do
 
   config :redix, url: System.fetch_env!("REDIS_URL")
 
+  config :sequin, Sequin.Mailer, adapter: Sequin.Swoosh.Adapters.Loops, api_key: System.get_env("LOOPS_API_KEY")
```
acco, line 140: We'll probably need a whole section on how to setup the mailer in self-hosted

As you'll see below, it's not like the Mailer will just work out of the box... I think that's something we can worry about if it's a problem for folks?

lib/sequin/accounts/user_notifier.ex
```diff
@@ -5,80 +5,47 @@ defmodule Sequin.Accounts.UserNotifier do
...
-  defp deliver(recipient, subject, body) do
+  defp deliver(recipient, subject, template_id, data_variables) do
     email =
       new()
       |> to(recipient)
       |> from({"Sequin", "support@sequinstream.com"})
       |> subject(subject)
-      |> text_body(body)
+      |> put_provider_option(:transactional_id, template_id)
+      |> put_provider_option(:data_variables, data_variables)
```
acco, line 15: This are specific properties for Loops

This is where we really "fork" and no longer have a generic Mailer. So, if someone wants to use SendGrid, they'll need to reconfigure this file to work with SendGrid's own bespoke provider options

This kinda makes sense broadly, as our email templates are on the provider, not in this application.

lib/sequin/accounts/user_notifier.ex
```diff
@@ -5,80 +5,47 @@ defmodule Sequin.Accounts.UserNotifier do
...
-    #{url}
-
-    If you didn't create an account with us, please ignore this.
-
-    ==============================
-    """)
+    deliver(user.email, "[Sequin] Please confirm your email", "cm0q0dm7f02pizz7zypxmjidb", %{
+      "user_email" => user.email,
+      "confirmation_url" => url
+    })
```
acco, line 29: Templates are now up on the provider

The template IDs are also hard-coded to our account. Another area of friction for self-hosted

lib/sequin/loops.ex
```diff
@@ -0,0 +1,108 @@
...
+    field :mailing_lists, map()
+  end
+
+  typedstruct module: TransactionalRequest do
+    field :email, String.t(), enforce: true
+    field :transactional_id, String.t(), enforce: true
+    field :add_to_audience, boolean()
+    field :data_variables, map()
+    field :attachments, list(map())
+  end
```
acco, line 28: "event requests" are what we might send for generic events: use signed up, posted their first webhook, etc. It's how we can trigger drips. 

transactional requests are things like confirm email, reset password, etc.

lib/sequin/loops_adapter.ex
```diff
@@ -0,0 +1,74 @@
+defmodule Sequin.Swoosh.Adapters.Loops do
```
acco, line 1: I just copied the SendGrid adapter into Cursor, pointed it to the API module, and said go

I'm not sure how much using Swoosh really buys us. Ultimately, we're just making a plain old HTTP request. Maybe it gets you a nice little mailbox for dev?

In theory, it should let you swap providers, but as demonstrated above that is not the real story.

acco (commented):
@RTLS low priority, take a look when convenient.

RTLS (approved):

PR #316, opened 2024-09-26
🐛 Skip replication tables if we lack permissions

RTLS:
It is possible that we receive replication messages for tables which our user does not have access to query.

In this PR, we simply skip these messages.

It would be preferable to alert the user. There are two levels of alerting when we do not have access:
1. A table is configured in the publication
2. A table is configured in the publication AND they have a consumer using that table

2 deserves sufficient alerting. 1 might not.

### DIFF
--- a/lib/sequin/extensions/replication.ex
+++ b/lib/sequin/extensions/replication.ex
@@ -264,9 +264,6 @@ defmodule Sequin.Extensions.Replication do
     and i.indisprimary;
     """
 
-    {:ok, %{rows: pk_rows}} = Postgres.query(conn, pk_query, [id])
-    primary_keys = List.flatten(pk_rows)
-
     # Query to get attnums
     # The Relation message does not include the attnums, so we need to query the pg_attribute
     # table to get them.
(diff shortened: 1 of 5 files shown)


RTLS:
* **#316** <a href="https://app.graphite.dev/github/pr/sequinstream/sequin/316?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

lib/sequin/extensions/replication.ex
```diff
@@ -329,23 +353,27 @@ defmodule Sequin.Extensions.Replication do
...
-      fields: data_tuple_to_fields(columns, msg.tuple_data),
-      trace_id: UUID.uuid4()
-    }
-
-    TracerServer.message_replicated(state.postgres_database, record)
-
-    %{state | accumulated_messages: [record | state.accumulated_messages]}
+    case Map.get(state.schemas, msg.relation_id) do
+      nil ->
+        state
```
acco, line 358: What if we did something a bit more assertive -- we can either keep a separate list in state of relation IDs for which we don't have permission. Or we can do something like `if Map.fetch!(state.schemas, msg.relation_id)[:process] do` or something

That way, we don't swallow any bugs where we didn't receive a relation message before receiving a WAL message

test/support/models/restricted_character.ex
```diff
@@ -0,0 +1,19 @@
+defmodule Sequin.Test.Support.Models.CharactersRestricted do
```
acco, line 1: Great

test/support/unboxed_repo/migrations/20240816005458_create_test_tables.exs
```diff
@@ -52,7 +53,29 @@ defmodule Sequin.Test.UnboxedRepo.Migrations.CreateTestTables do
...
+                CREATE ROLE replication_user WITH REPLICATION LOGIN PASSWORD 'postgres';
+              END IF;
+            END
+            $$;
+            """,
+            "DROP ROLE IF EXISTS replication_user"
+
+    execute "REVOKE USAGE ON SCHEMA restricted FROM replication_user;", ""
+
+    execute "create publication characters_publication for table characters, characters_ident_full, characters_multi_pk, characters_detailed, restricted.characters_restricted",
```
acco, line 78: 💵

RTLS:
Looks like using relid bypasses the permissions issue!

PR #372, opened 2024-10-10
Auto-create replication slots for Neon databases

acco:
Neon databases cull their replication slots after 45min of inactivity. This commit ensures we recreate them, should the db have gone dormant.

Long term, we want to keep track of the LSN that we last ack'd for the rep slot, so we can also rewind the slot on create. And probably turn this into an opt-in option for dev databases.

For now, this level of inactivity is only expected for dev dbs.


acco:
* **#372** <a href="https://app.graphite.dev/github/pr/sequinstream/sequin/372?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

PR #377, opened 2024-10-11
:bug: Fix issue with null UUIDs

acco:
Postgrex does not automatically convert string UUIDs to binary before querying, nor convert to UTF8 string on load. This means we have to use column types to determine when to do this coercion. However, we were improperly handling situations where the value of a UUID field is null.

s/o to @aglassman for both reporting and identifying the root cause of this issue! :pray:


acco:
* **#377** <a href="https://app.graphite.dev/github/pr/sequinstream/sequin/377?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

RTLS (approved):

PR #389, opened 2024-10-14
feat: add account user and pending invites table

yordis:


assets/svelte/settings/AccountSettings.svelte
```diff
@@ -93,59 +113,163 @@
...
-        >
-          Delete Account
+    <div class="bg-card text-card-foreground rounded-lg border shadow-sm">
+      <div class="flex p-6 justify-between">
+        <h3 class="text-lg font-semibold leading-none tracking-tight">
+          Invitations
+        </h3>
+        <Button>
+          <Plus class="h-4 w-4 mr-2" />
+          Invite user
```
acco, line 215: Hide when there are 0 pending users

yordis: How could you invite new users, then?

yordis: done

yordis:
@RTLS share some love over here btw, I will continue the work in the modal after this one

PR #437, opened 2024-10-24
Improve self-hosted hosted telemetry

thisisgoldman:


lib/sequin_web/components/layouts/root.html.heex
```diff
@@ -19,7 +19,13 @@
   </head>
   <body
     class="bg-white antialiased"
-    data-ph-token={Application.get_env(:sequin, :frontend_posthog_api_key)}
```
thisisgoldman, line 22: If telemetry is disabled, then no API key is returned to the front end

RTLS (approved):

PR #484, opened 2024-11-05
✨ Show last_committed stats in databases/show

RTLS:


RTLS:
* **#484** <a href="https://app.graphite.dev/github/pr/sequinstream/sequin/484?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

RTLS:
This method of measuring replication lag is imperfect and is not an improvement to the UX in my opinion.

`pg_stat_replication` gives us `flush_lag` for active slots. When there is constant activity, `flush_lag` is a good measure of how long transactions take from committing on the primary to acknowledgment by Sequin.

When there is no activity on the primary for tables in the publication slot, the `flush_lag` grows linearly with time. Eventually, Postgres returns null for the value which we can use as an indicator that we are caught up. The transtion to null seems to occur after 13 minutes.

We can also measure the delay in bytes, which is supposedly more accurate / less prone to errors due to inactivity, but on my system it also grows linearly with time

```
-- Monitor lag in bytes
SELECT 
    slot_name,
    pg_size_pretty(pg_current_wal_lsn() - restart_lsn) as lag_bytes,
FROM pg_replication_slots
WHERE slot_type = 'logical';
```

We can opt to only show replication lag if we are currently ingesting messages from upstream. But this does not solve whether we have accurate stats; just whether we show inaccurate stats to our users.

The most robust way to measure replication lag is to intentionally set a row in the primary database with a timestamp then measure when we receive that row in the replication slot. Is it worth doing this work?

PR #493, opened 2024-11-06
Use original Wistia embed without `<iframe>`

ricardonunez-io:
Hi there! Found the issue with the embed, the HTML given from the Wistia site is not valid JSX, so the markdown parsing broke. 

I fixed it in this PR to replace the `<iframe>`. Just the style props were broken (since HTML has strings for style props instead of objects which JSX takes), and the `onLoad` props which had a string instead of Javascript.

Thank you!


acco:
Thank you @ricardonunez-io , much appreciated!

PR #572, opened 2024-11-18
Set SECRET_KEY_BASE length to 64 bytes

mdminhazulhaque:
The doc mistakenly mentions the length of SECRET_KEY_BASE as 32 byte long. It should 64 byte long.


acco:
Thanks, @mdminhazulhaque ! Much appreciated

PR #671, opened 2024-11-25
✨ /version for version info in self hosted

RTLS:


lib/sequin_web/router.ex
```diff
@@ -70,19 +78,6 @@ defmodule SequinWeb.Router do
     end
   end
 
-  scope "/", SequinWeb do
```
RTLS, line 73: I combined all the routes that do not require authentication or lack of authentication into one scope

PR #682, opened 2024-11-27
Set SECRET_KEY_BASE length to 64 bytes

mdminhazulhaque:
Hello @acco 

Similar to #572 suggesting a change in the doc.

Thanks.


acco:
@mdminhazulhaque Thanks again! 🙏

PR #700, opened 2024-12-08
[Docs] Fix example configuration

harrybuisman:
Docker container stops when using the example configuration from the docs due to the incorrect destination type

```
** (FunctionClauseError) no function clause matching in Sequin.YamlLoader.parse_sink/2 
```


RTLS:
Thank you @harrybuisman !

RTLS:
Thank you @harrybuisman !

acco:
@harrybuisman Harry, thank you! Good catch.

Let us know if you need help with anything

PR #708, opened 2024-12-11
feat: add nats sink

yordis:


assets/svelte/components/SinkCardNatsJetstream.svelte
```diff
@@ -0,0 +1,56 @@
...
+        <div class="mt-2">
+          <span
+            class="font-mono bg-slate-50 pl-1 pr-4 py-1 border border-slate-100 rounded-md whitespace-nowrap"
+            >{consumer.sink.port}</span
+          >
+        </div>
+      </div>
+
+      <div>
+        <span class="text-sm text-gray-500">Stream name</span>
```
acco, line 36: Rename

assets/svelte/consumers/SinkNatsJetstreamForm.svelte
acco, line 1: Looking great!

lib/sequin/consumers/nats_jetstream_sink.ex
```diff
@@ -0,0 +1,177 @@
+defmodule Sequin.Consumers.NatsJetstreamSink do
```
acco, line 1: per Discord, rename to `NatsSink`

lib/sequin/consumers/nats_jetstream_sink.ex
```diff
@@ -0,0 +1,177 @@
...
+  use TypedEctoSchema
+
+  import Ecto.Changeset
+
+  alias Sequin.Encrypted.Binary
+
+  @derive {Jason.Encoder, only: [:host, :port, :stream_name]}
+  @primary_key false
+  typed_embedded_schema do
+    field :type, Ecto.Enum, values: [:nats_jetstream], default: :nats_jetstream
```
acco, line 13: ditto rename here

lib/sequin/consumers/nats_jetstream_sink.ex
```diff
@@ -0,0 +1,177 @@
...
+    |> validate_required([:host, :port, :stream_name])
+    |> validate_number(:port, greater_than: 0, less_than: 65_536)
+    |> validate_length(:stream_name, max: 255)
+    |> validate_auth_credentials()
+    |> validate_tls_config()
+    |> validate_nats_host()
+    |> put_new_connection_id()
+  end
+
+  defp validate_auth_credentials(changeset) do
```
acco, line 54: nice

lib/sequin/consumers/nats_jetstream_sink.ex
```diff
@@ -0,0 +1,177 @@
...
+        add_error(changeset, :tls_cert_file, "required when key file is provided")
+
+      true ->
+        changeset
+    end
+  end
+
+  defp validate_nats_host(changeset) do
+    if prod_env?() and not self_hosted?() do
+      case System.fetch_env("NATS_URL") do
```
acco, line 108: Very fun - Claude probably copied this over from our Redis sink :D Not necessary for this one

lib/sequin/consumers/nats_jetstream_sink.ex
```diff
@@ -0,0 +1,177 @@
...
+      changeset
+    end
+  end
+
+  defp put_new_connection_id(changeset) do
+    case get_field(changeset, :connection_id) do
+      nil -> put_change(changeset, :connection_id, Ecto.UUID.generate())
+      _ -> changeset
+    end
+  end
```
acco, line 132: Don't need a connection_id for NATS, this was just for Kafka/:brod

lib/sequin/consumers_runtime/nats_jetstream_pipeline.ex
```diff
@@ -0,0 +1,128 @@
...
+
+    Broadway.start_link(__MODULE__,
+      name: via_tuple(consumer.id),
+      producer: [
+        module: {producer, [consumer: consumer, test_pid: test_pid]}
+      ],
+      processors: [
+        default: [
+          concurrency: consumer.max_waiting,
+          max_demand: 10
```
acco, line 34: This is the batch size for payloads sent to NATS. Either 10 or 100 is good. (I assume there is some benefit in sending batches to NATS - though not sure about the client specifics?)

acco: Ah, looks like we don't batch! In that case, unless we have an idempotency key on the NATS side, we should just set this to 1 - that way, we don't get into a situation where 5 out of a batch of 10 were written before we crashed

lib/sequin/consumers_runtime/nats_jetstream_pipeline.ex
```diff
@@ -0,0 +1,128 @@
...
+      service: :nats_jetstream,
+      code: "publish_error",
+      message: "NATS JetStream publish failed",
+      details: %{error: error}
+    )
+  end
+
+  defp setup_allowances(nil), do: :ok
+
+  defp setup_allowances(test_pid) do
```
acco, line 125: You'll need this only if you decide to stub NATS for test (probably?)

lib/sequin/consumers_runtime/nats_jetstream_pipeline.ex
```diff
@@ -0,0 +1,128 @@
...
+          )
+        end)
+
+        Broadway.Message.failed(message, reason)
+    end
+  end
+
+  defp publish_messages(conn, stream_name, messages) do
+    for msg <- messages do
+      :ok = Gnat.pub(conn, stream_name, Jason.encode!(msg.data))
```
acco, line 106: I assume you will use Nats.Client

lib/sequin/consumers_runtime/nats_jetstream_pipeline.ex
```diff
@@ -0,0 +1,128 @@
...
+      ],
+      processors: [
+        default: [
+          concurrency: consumer.max_waiting,
+          max_demand: 10
+        ]
+      ],
+      context: %{
+        consumer: consumer,
+        nats_conn: NatsJetstreamSink.connection_opts(consumer.sink),
```
acco, line 39: Looks like we won't need this, we'll just pass the SinkConsumer to Nats.Client

lib/sequin/nats/client.ex
```diff
@@ -0,0 +1,99 @@
+defmodule Sequin.Nats.Client do
```
acco, line 1: Very ncie

lib/sequin/nats/nats.ex
```diff
@@ -0,0 +1,37 @@
...
+  end
+
+  @spec test_connection(NatsJetstreamSink.t()) :: :ok | {:error, Error.t()}
+  def test_connection(%NatsJetstreamSink{} = sink) do
+    impl().test_connection(sink)
+  end
+
+  defp impl do
+    Application.get_env(:sequin, :nats_module, Sequin.Nats.Client)
+  end
```
acco, line 36: 👍

yordis: why we couldn't use `compiled_env` here?

lib/sequin/consumers/nats_sink.ex
```diff
@@ -0,0 +1,178 @@
...
+      :host,
+      :port,
+      :username,
+      :password,
+      :token,
+      :nkey_seed,
+      :jwt,
+      :tls,
+      :tls_cert_file,
+      :tls_key_file,
```
yordis, line 41: how to manage this?

assets/svelte/consumers/SinkNatsForm.svelte
```diff
@@ -0,0 +1,262 @@
...
+          <p class="text-destructive text-sm">{errors.sink.tls_cert_file}</p>
+        {/if}
+      </div>
+
+      <div class="space-y-2">
+        <Label for="tls_key_file">TLS key file</Label>
+        <Input
+          id="tls_key_file"
+          bind:value={form.sink.tls_key_file}
+          placeholder="/path/to/key.pem"
```
acco, line 254: I assume they'll just paste a blob in? Or are you thinking upload?

acco: Ah, i see your comment below

lib/sequin/consumers/nats_sink.ex
acco: Just store as an encrypted text field in Postgres?

lib/sequin/consumers/nats_sink.ex
```diff
@@ -0,0 +1,178 @@
...
+
+      true ->
+        changeset
+    end
+  end
+
+  defp validate_nats_host(changeset) do
+    if prod_env?() and not self_hosted?() do
+      # TODO: why this? and why to read from env?
+      case System.fetch_env("NATS_URL") do
```
acco, line 109: I left a comment in my last PR review – we had this gate to prevent users from accessing our internal/private redis instance on cloud :D 

We do not need for NATS – so you can remove this whole function/clause

lib/sequin/consumers/nats_sink.ex
```diff
@@ -0,0 +1,178 @@
...
+
+        :error ->
+          changeset
+      end
+    else
+      changeset
+    end
+  end
+
+  defp put_new_connection_id(changeset) do
```
acco, line 128: Again, we needed for :brod + ConnectionCache, may not need here

lib/sequin/consumers_runtime/nats_pipeline.ex
```diff
@@ -0,0 +1,110 @@
...
+      message: "NATS publish failed",
+      details: %{error: error}
+    )
+  end
+
+  defp setup_allowances(nil), do: :ok
+
+  defp setup_allowances(test_pid) do
+    Req.Test.allow(Sequin.Aws.HttpClient, test_pid, self())
+  end
```
acco, line 109: you can del

lib/sequin/consumers_runtime/nats_pipeline.ex
```diff
@@ -0,0 +1,110 @@
...
+
+    Broadway.start_link(__MODULE__,
+      name: via_tuple(consumer.id),
+      producer: [
+        module: {producer, [consumer: consumer, test_pid: test_pid]}
+      ],
+      processors: [
+        default: [
+          concurrency: consumer.max_waiting,
+          max_demand: 10
```
acco, line 31: Use max_demand: 1 (1 message sent to NATS at time) unless we can set idempotency key

The idempotency key could be the consumer_record/consumer_event's `id`

An even better one is to use a combination of the record's `seq` + `record_pks` -- those are guaranteed to be a unique combo. `record_pks` is a `text[]` and `seq` is a bigint though - so you'd have to figure out how to combine them for a NATS idempotency key

acco: I see your comment below/implementation of idempotency below!

lib/sequin/nats/client.ex
```diff
@@ -0,0 +1,102 @@
...
+    with {:ok, connection} <- ConnectionCache.connection(sink) do
+      Enum.reduce_while(messages, :ok, fn
+        %ConsumerEvent{} = message, :ok ->
+          payload = %{
+            record: message.data.record,
+            metadata: message.data.metadata
+          }
+
+          # TODO: handle idempotency, verify this is correct
+          opts = [headers: [{"Nats-Msg-Id", message.id}]]
```
acco, line 23: This works! I think using the `id` is a good starting place. I can't think of a good failure scenario that warrants something more complex, like using `seq` + `record_pks`

lib/sequin/nats/client.ex
```diff
@@ -0,0 +1,102 @@
...
+        {:ok, %{body: info}} -> {:ok, info}
+        {:error, error} -> {:error, to_sequin_error(error)}
+      end
+    end
+  end
+
+  @impl Sequin.Nats
+  def test_connection(%NatsSink{} = sink) do
+    with :ok <-
+           NetworkUtils.test_tcp_reachability(sink.host, sink.port, sink.tls, :timer.seconds(10)),
```
acco, line 65: The third argument is actually `ipv6` (boolean). It's possible we messed this up elsewhere.

We have [check_ipv6](https://github.com/sequinstream/sequin/blob/master/lib/sequin/network_utils.ex#L9-L21) that determines if a host/port needs ipv6 to connect.

We're not in the practice of this yet, but probably should: storing whether a connection should use ipv6 (by using check_ipv6 on save). Alternatively, we can just check before starting a connection.

**However**, we can do this as a fast-follow post launch

lib/sequin/nats/nats.ex
acco: @yordis We can!

lib/sequin/consumers/nats_jetstream_sink.ex
yordis: Since this is so critical, I wonder if you should create a `validate_managed_service` or something like that,

I am gonna remove it but if that changes in the future ... you are in it for trouble hehe

assets/svelte/types/consumer.ts
```diff
@@ -67,6 +67,16 @@ export type RedisConsumer = BaseConsumer & {
   };
 };
 
+// NATS specific sink
```
yordis, line 70: image

acco:
* **#708** <a href="https://app.graphite.dev/github/pr/sequinstream/sequin/708?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/sequinstream/sequin/708?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

assets/svelte/types/consumer.ts
RTLS: 😂

acco:
Merged, thanks a million Yordis!

24154e6a248ca04c24c3e89c8b8662a64752ff85
0b8d8c0bb87aeb24ed8f6ec1d10e1b87a3aa76c3
aba3fefa4bd9d0b5da3584217414a748f37fcac1

PR #728, opened 2024-12-14
feat: add rabbitmq sink

yordis:


lib/sequin/consumers_runtime/rabbit_mq_pipeline.ex
```diff
@@ -0,0 +1,104 @@
...
+      {:error, error} ->
+        reason =
+          Error.service(
+            service: :rabbitmq,
+            code: "publish_error",
+            message: "RabbitMQ publish failed",
+            # TODO: I ran into an issue encoding the error message to JSON
+            # maybe, I should check `is_struct(error, ServiceError)`?
+            # otherwise it is dangerous to propogate errors
+            details: %{error: error}
```
acco, line 81: Yeah we can import a guard and only propagate if it's a Sequin Error? Or `inspect/1` if it's not?

https://github.com/sequinstream/sequin/blob/master/lib/sequin/error.ex#L213

lib/sequin/rabbit_mq/client.ex
```diff
@@ -0,0 +1,83 @@
...
+    end
+  end
+
+  @impl Sequin.RabbitMq
+  def test_connection(%RabbitMqSink{} = sink) do
+    with :ok <- NetworkUtils.test_tcp_reachability(sink.host, sink.port, RabbitMqSink.ipv6?(sink), :timer.seconds(10)),
+         {:ok, _connection} <- ConnectionCache.connection(sink) do
+      # TODO: figure out how to test connection
+      # Check https://www.rabbitmq.com/docs/troubleshooting-networking
+      :ok
```
acco, line 34: We can maybe do a `message_count` on the queue (if cheap) or do something silly like ack a non-existent message

lib/sequin/consumers_runtime/rabbit_mq_pipeline.ex
yordis: No inspects, we could leak out internals of the managed cloud service

PR #752, opened 2024-12-20
chore: add iex local file

yordis:


.iex.exs
```diff
@@ -1,7 +1,11 @@
+IEx.configure(auto_reload: true)
```
yordis, line 1: elixir 1.18+ will use this!

RTLS (approved):

PR #776, opened 2024-12-23
🚧 Add ConsumerMessageStore

acco:


acco:
* **#776** <a href="https://app.graphite.dev/github/pr/sequinstream/sequin/776?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/sequinstream/sequin/776?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

lib/sequin/consumers/consumer_event.ex
```diff
@@ -39,6 +39,10 @@ defmodule Sequin.Consumers.ConsumerEvent do
 
     embeds_one :data, ConsumerEventData
```
acco, line 40: TODO: We need to add group_id to ConsumerEvent. Wise to get that into `main`, then rebase?

We're re-visiting the way we pull messages, set not visible, ack them. So a good time to do it.

lib/sequin/consumers/consumer_event.ex
```diff
@@ -39,6 +39,10 @@ defmodule Sequin.Consumers.ConsumerEvent do
...
 
+    # For ConsumerMessageStore
+    field :flushed_at, :utc_datetime_usec, virtual: true
+    field :dirty, :boolean, virtual: true
```
acco, line 44: `flushed_at` - used to determine when we last flushed the message to Postgres. Starts off as `nil`
`dirty` - we need to flip every time we change the message in ConsumerMessageStore

We don't flush stuff unless it's dirty. So `flushed_at` is just so we can order and prioritize which dirty things to flush.

acco: Also down for `last_flushed_at`

lib/sequin/consumers/consumer_record.ex
```diff
@@ -37,6 +37,12 @@ defmodule Sequin.Consumers.ConsumerRecord do
 
     field :data, :map, virtual: true
 
+    # For ConsumerMessageStore
+    field :flushed_at, :utc_datetime_usec, virtual: true
+    field :dirty, :boolean, virtual: true
+    # Used to track if the record has been deleted from MessageHandler -> Store
+    field :deleted, :boolean, virtual: true
```
acco, line 44: Right now, the behavior in MessageHandler is:

1. If we are processing a `:record` consumer
2. And we get a `delete` message
3. Go delete entries for that consumer in `consumer_records`

It's kind of unnecessary? Because we're racing the record out the door? 

So I put this here so that we could track deleted records in CMS. But you'll note that we don't actually use this.

I think the better solution later is to support delete markers. Perhaps to keep things simple for now, we don't do anything when we receive deletes

lib/sequin/consumers/consumers.ex
```diff
@@ -367,6 +367,36 @@ defmodule Sequin.Consumers do
...
+      end)
+
+    # insert_all expects a plain outer-map, but struct embeds
+    {count, _} =
+      Repo.insert_all(
+        ConsumerEvent,
+        events,
+        on_conflict: {:replace, [:state, :updated_at, :deliver_count, :last_delivered_at, :not_visible_until]},
+        conflict_target: [:consumer_id, :commit_lsn]
+      )
```
acco, line 387: upsert_consumer_events and upsert_consumer_records are very similar to the `insert_*` functions MessageHandler currently calls

Excep the `on_conflict` replace - we overwrite all the mutable columns

Now we're doing all the calculation around `deliver_count` and `not_visible_until` in CMS. Postgres becomes dumber (this logic used to happen in our `receive` SQL calls). So we want to do a "plain" upsert here

We'll call this as part of flush

lib/sequin/consumers_runtime/consumer_message_store.ex
```diff
@@ -0,0 +1,360 @@
...
+  alias Sequin.Consumers.ConsumerRecord
+  alias Sequin.Consumers.SinkConsumer
+  alias Sequin.Error
+
+  require Logger
+
+  defguardp event_messages?(state) when state.consumer.message_kind == :event
+  defguardp record_messages?(state) when state.consumer.message_kind == :record
+
+  defmodule State do
```
acco, line 32: We can either do the State module for pure functions (here)

Or, we can break this out into a separate module. Maybe as it grows? 

I think it makes sense to pass `State` struct around?

lib/sequin/consumers_runtime/consumer_message_store.ex
```diff
@@ -0,0 +1,360 @@
...
+    alias Sequin.Consumers.ConsumerEvent
+    alias Sequin.Consumers.ConsumerRecord
+
+    defguardp event_messages?(state) when state.consumer.message_kind == :event
+    defguardp record_messages?(state) when state.consumer.message_kind == :record
+
+    typedstruct do
+      field :consumer_id, String.t()
+      field :consumer, SinkConsumer.t()
+      field :messages, %{[String.t()] => ConsumerRecord.t() | ConsumerEvent.t()}
```
acco, line 45: Started with a plain list, but one operation in particular pushed to map: when we had to merge in messages, some updates some inserts

lib/sequin/consumers_runtime/consumer_message_store.ex
```diff
@@ -0,0 +1,360 @@
...
+                   # Preserve delivery-related fields
+                   deliver_count: existing.deliver_count,
+                   last_delivered_at: existing.last_delivered_at,
+                   not_visible_until: existing.not_visible_until,
+                   ack_id: existing.ack_id,
+                   # Mark as dirty for next flush
+                   dirty: true
+               }}
+          end
+        end)
```
acco: This mirrors this logic:

https://github.com/sequinstream/sequin/blob/main/lib/sequin/consumers/consumers.ex#L482-L540

lib/sequin/consumers_runtime/consumer_message_store.ex
```diff
@@ -0,0 +1,363 @@
...
+                   # We update commit_lsn and seq to match the latest version of the message
+                   state: state,
+                   commit_lsn: updated.commit_lsn,
+                   seq: updated.seq,
+
+                   # Mark as dirty for next flush
+                   dirty: true
+               }}
+          end
+        end)
```
acco, line 100: This mirrors this logic:

https://github.com/sequinstream/sequin/blob/main/lib/sequin/consumers/consumers.ex#L482-L540

lib/sequin/consumers_runtime/consumer_message_store.ex
```diff
@@ -0,0 +1,350 @@
...
+              state: :delivered,
+              dirty: true
+          }
+        end)
+
+      # Replace old messages with updated ones in state
+      state = update_messages(state, messages)
+
+      {state, messages}
+    end
```
acco, line 150: We want same logic that we have in receive_consumer_events and receive_record_events. Double check me!

Known change is that we order by `seq` (instead of ID, which we do now).

Claude thinks we're equivalent

lib/sequin/consumers_runtime/consumer_message_store.ex
```diff
@@ -0,0 +1,350 @@
...
+        if not_visible_until = ack_ids_with_not_visible_until[msg.ack_id] do
+          msg = %{msg | not_visible_until: not_visible_until, state: :available, dirty: true}
+          {record_pks, msg}
+        else
+          {record_pks, msg}
+        end
+      end)
+
+    {:reply, :ok, %{state | messages: updated_messages}}
+  end
```
acco, line 301: Same as:
https://github.com/sequinstream/sequin/blob/main/lib/sequin/consumers/consumers.ex#L1041-L1069

lib/sequin/consumers_runtime/consumer_message_store.ex
```diff
@@ -0,0 +1,350 @@
...
+      {_deletes, upserts} = messages |> Map.values() |> Enum.split_with(& &1.deleted)
+
+      # Split messages into updates and inserts based on existing record_pks
+      {updates, inserts} =
+        Enum.split_with(upserts, fn msg ->
+          Map.has_key?(state.messages, msg.record_pks)
+        end)
+
+      updates_by_pk = Map.new(updates, &{&1.record_pks, &1})
+      inserts_by_pk = Map.new(inserts, &{&1.record_pks, &1})
```
RTLS, line 65: Do we have to worry about multiple updates or inserts for a `record_pks` here? if there are > 1 we will lose some in this map

lib/sequin/consumers_runtime/consumer_message_store.ex
```diff
@@ -0,0 +1,350 @@
...
+
+    {:reply, :ok, state}
+  end
+
+  def handle_call({:produce, count}, _from, %State{} = state) do
+    deliverable_messages = State.deliverable_messages(state, count)
+    {state, messages} = State.deliver_messages(state, deliverable_messages)
+
+    {:reply, {:ok, messages}, state}
+  end
```
RTLS, line 268: where do we load records from the customer postgres? ie. `Consumers.put_source_data`

lib/sequin/consumers_runtime/consumer_message_store.ex
```diff
@@ -0,0 +1,350 @@
...
+
+  def handle_call({:ack, ack_ids}, _from, state) do
+    # Delete from database right away
+    # TODO: We need to respect pending_redelivery on consumer records
+    # TODO: We can greatly simplify this call now by just deleting events and records
+    Consumers.ack_messages(state.consumer, ack_ids)
```
RTLS, line 274: we should probably move this to the client process in `def ack` before the `GenServer.call`

lib/sequin/consumers_runtime/consumer_message_store.ex
```diff
@@ -0,0 +1,350 @@
...
+    # Update messages with matching ack_ids to set their not_visible_until and state
+    updated_messages =
+      Map.new(state.messages, fn {record_pks, msg} ->
+        if not_visible_until = ack_ids_with_not_visible_until[msg.ack_id] do
+          msg = %{msg | not_visible_until: not_visible_until, state: :available, dirty: true}
+          {record_pks, msg}
+        else
+          {record_pks, msg}
+        end
+      end)
```
RTLS, line 298: minor but probably better to iterate over `ack_ids_with_not_visible_until` than `state.messages` - `state.messages` might be very large

lib/sequin/consumers_runtime/consumer_message_store.ex
acco: @RTLS Good call - that's missing atm

lib/sequin/consumers_runtime/consumer_message_store.ex
acco: @RTLS That makes sense

lib/sequin/consumers_runtime/consumer_message_store.ex
acco: Yes good point @RTLS

lib/sequin/consumers_runtime/consumer_message_store.ex
acco: @RTLS It's there that we can do the "leaf idempotency" as well

PR #778, opened 2024-12-23
✅ Add ConsumerMessageStore GenServer test

RTLS:


RTLS:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/sequinstream/sequin/778?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#778** <a href="https://app.graphite.dev/github/pr/sequinstream/sequin/778?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/sequinstream/sequin/778?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#776** <a href="https://app.graphite.dev/github/pr/sequinstream/sequin/776?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

lib/sequin/consumers_runtime/consumer_message_store.ex
```diff
@@ -264,18 +288,33 @@ defmodule Sequin.ConsumersRuntime.ConsumerMessageStore do
     deliverable_messages = State.deliverable_messages(state, count)
     {state, messages} = State.deliver_messages(state, deliverable_messages)
 
-    {:reply, {:ok, messages}, state}
+    if state.consumer.message_kind == :record do
+      with {:ok, messages} <- Consumers.put_source_data(state.consumer, messages) do
```
acco, line 292: Seeing this here -- given the requirements shift, at this point, can't we (shouldn't we) just use the full record coming through the slot? Then as new messages come in, we stomp over the ones in the store. (Means adding a column for the record in the table)

This is a for-later. We should probably get this in changing ConsumerRecord and events as little as possible. Then merge them into a more generic WAL event, storing everything on there (and discarding changes for records on delivery).

lib/sequin/consumers_runtime/consumer_message_store.ex
```diff
@@ -264,18 +288,33 @@ defmodule Sequin.ConsumersRuntime.ConsumerMessageStore do
...
-    # TODO: We need to respect pending_redelivery on consumer records
-    # TODO: We can greatly simplify this call now by just deleting events and records
-    Consumers.ack_messages(state.consumer, ack_ids)
-
     # Delete from in-memory store
     id_set = MapSet.new(ack_ids)
 
+    # TODO: optimize this with a secondary index on ack_id?
+    # with 100k messages, this takes 230 ms
+    # with 1M messages, this takes 3 seconds
```
acco, line 317: Probably a good idea

acco: Possible future optimization would be to have "active" messages in their own data structure, so we know it gets to a fixed size (max ack pending)

Though if we've gotten to a point where we've accumulated 1M messages in the store, something is probably off. Back pressure in the system is needed

test/sequin/consumer_message_store_test.exs
```diff
@@ -0,0 +1,134 @@
...
+  describe "Event Sink Consumer - message handling" do
+    setup do
+      consumer = ConsumersFactory.insert_sink_consumer!(message_kind: :event)
+
+      start_supervised!({ConsumerMessageStore, consumer_id: consumer.id, test_pid: self()})
+
+      %{consumer: consumer}
+    end
+
+    test "puts, delivers, nacks, and acks messages", %{consumer: consumer} do
```
acco, line 23: great test!

lib/sequin/consumers_runtime/consumer_message_store.ex
RTLS: Definitely agree. We will have to specifically optimize for back-pressure situations at some point. I figure we should focus on correctness first then come back to optimizations like this.

lib/sequin/consumers_runtime/consumer_message_store.ex
RTLS: I strongly agree we should avoid going to the customer database in ConsumerMessageStore- that is guaranteed to be a bottleneck!

test/sequin/consumer_message_store_test.exs
RTLS: In terms of more specific state tests such as:
* delivered records are not visible some time
* nacked records respect not visible until
* n/acking 1 of many delivered records effects the correct record
etc.

we should test those on the state module directly. these GenServer tests are just verifying the glue code

PR #779, opened 2024-12-24
✨ ConsumerIdempotency

RTLS:


lib/sequin/consumers_runtime/consumer_idempotency.ex
```diff
@@ -0,0 +1,62 @@
+defmodule Sequin.ConsumersRuntime.ConsumerIdempotency do
+  @moduledoc """
+  Handles idempotency for consumers at "the leaf", i.e. at the delivery time.
+
+  After successfully delivering messages to a sink destination, the first thing we do
+  is mark the messages as delivered via this module.
+
+  The last thing we do before deliverying a message is to check if we have previously marked
+  the message as delivered. If so, we skip deliverying it and consider it delivered.
```
acco, line 9: new fave word

lib/sequin/consumers_runtime/consumer_idempotency.ex
```diff
@@ -0,0 +1,62 @@
...
+
+  We rely on the ConsumerMessageStore to trim the set of ackknowledted messages
```
acco, line 11: nvm, new fave word

acco: you know, it makes the comment authentic - i know it was hand written

lib/sequin/consumers_runtime/consumer_idempotency.ex
```diff
@@ -0,0 +1,62 @@
...
+  @type seq :: integer()
+  @type message :: ConsumerRecord.t() | ConsumerEvent.t()
+
+  @spec mark_messages_delivered(consumer_id(), [message()]) :: :ok | {:error, Error.t()}
+  def mark_messages_delivered(_, []), do: :ok
+
+  def mark_messages_delivered(consumer_id, messages) do
+    messages
+    |> Enum.map(fn message -> ["ZADD", delivered_key(consumer_id), message.seq, message.seq] end)
+    |> Redis.pipeline()
```
acco, line 30: One thing I'm still weary of with Redis clustering is I'm not sure if it affects these multi calls. We should test.

lib/sequin/consumers_runtime/consumer_producer.ex
```diff
@@ -62,13 +66,39 @@ defmodule Sequin.ConsumersRuntime.ConsumerProducer do
...
+
+      {:ok, lsn} ->
+        ConsumerIdempotency.trim(state.consumer.id, lsn)
+
+      {:error, error} ->
+        Logger.error("Error trimming idempotency seqs", error: error)
+    end
+
+    {:noreply, [], schedule_trim_idempotency(state)}
+  end
```
acco, line 83: ah yes - reasonable place to put it, given the sets are by consumer_id (which makes sense) 👍

lib/sequin/consumers_runtime/consumer_producer.ex
```diff
@@ -62,13 +66,39 @@ defmodule Sequin.ConsumersRuntime.ConsumerProducer do
...
 
+    {:ok, delievered_seqs} = ConsumerIdempotency.delivered_messages(state.consumer.id, messages)
+    messages = Enum.reject(messages, &(&1.seq in delievered_seqs))
+
+    unless delievered_seqs == [] do
+      Logger.warning(
+        "Received #{length(delievered_seqs)} messages for consumer #{state.consumer.id} that have already been delivered",
+        seqs: delievered_seqs
+      )
+    end
```
acco, line 100: great

lib/sequin/consumers_runtime/consumer_idempotency.ex
RTLS: 😂

girishramnani:
A very late design question - why not use the sink to figure out if a message was already delivered?

acco:
@girishramnani Good question! Not every sink makes this easy: for example, we can't ask SQS which messages have already been delivered. For the webhook sink, once a message has been delivered, it disappears.

Redis gives us a canonical place to do idempotency

girishramnani:
Hmm gotcha

girishramnani:
Not sure how important triming is but I would recommend Bloom filters if sorting isn’t important https://redis.io/docs/latest/develop/data-types/probabilistic/bloom-filter/
Better space complexity

acco:
@girishramnani I wish we could use a bloom filter! Unfortunately, if it gives us a false-positive "possibly in set" (even if it isn't) - that would result in us not delivering a message, which is bad.

girishramnani:
Makes sense

PR #793, opened 2024-12-31
fix: set 'self_hosted: true' in dev.exs to hide 'Use Localhost' toggle

girishramnani:
fix: set 'self_hosted: true' in dev.exs to hide 'Use Localhost' toggle


girishramnani:
Closing as we want to keep this false

Issue #688, opened 2024-12-04
Add ability to ack/delete messages in the console


acco:
When in the console, on a sink's "Messages" tab, add the ability to ack/delete individual messages.

girishramnani:
Assuming this would be different for each sink’s it should be part of the sink’s interface? Haven’t seen if there is a behaviour defined for sinks. If not then we should design one.

acco:
@girishramnani This is actually the same for all sinks! You'd just call `Consumers.ack_messages(consumer, [message.ack_id])`

You could add this button in the Message slide-out. Near this other button, which would be styled the same way:

https://github.com/sequinstream/sequin/blob/master/assets/svelte/consumers/ShowMessages.svelte#L548-L575

girishramnani:
Awesome, I would like to work on this.
Just to get ahead. I should add a test too right?

PR #796, opened 2025-01-01
feature: Support acknowledging messages in a sink

girishramnani:
feature: Support acknowledging messages in a sink

Closes https://github.com/sequinstream/sequin/issues/688

### DIFF
--- a/assets/svelte/consumers/ShowMessages.svelte
+++ b/assets/svelte/consumers/ShowMessages.svelte
@@ -285,6 +285,9 @@
         // Handle error (e.g., show a toast notification)
         console.error("Failed to acknowledge:", reply.error);
       }
+      if (reply.ok) {
+        selectedMessage = reply.updated_message;
+      }
     });
   }
 
@@ -593,9 +596,14 @@
                     <h3 class="text-lg font-semibold mb-2">Acknowledge</h3>
                     <div class="bg-gray-50 p-4 rounded-lg space-y-4">
                       <p class="text-sm text-gray-700">
-                        Acknowledge this message to remove it from the sink's
-                        delivery queue.
+                        {#if alreadyAcked}
+                          This message has been acknowledged.
+                        {:else}
+                          Acknowledge this message to remove it from the sink's
+                          delivery queue.
+                        {/if}
                       </p>
+
                       <Button
                         variant="outline"
                         size="sm"
@@ -604,7 +612,7 @@
                           alreadyAcked = true;
                         }}
                         class="flex items-center space-x-2"
-                        disabled={alreadyAcked}
+                        hidden={alreadyAcked}
                       >
                         <RotateCw class="h-4 w-4" />
                         <span>Acknowledge</span>
--- a/assets/svelte/consumers/ShowMessages.svelte
+++ b/assets/svelte/consumers/ShowMessages.svelte
@@ -255,6 +255,25 @@
 
   function resetMessageVisibility(messageId) {
     live.pushEvent(
+      "reset_message_visibility",
+      { message_id: messageId },
+      (reply) => {
+        resettingMessageVisibility = false;
+
+        if (reply.error) {
+          // Handle error (e.g., show a toast notification)
+          console.error("Failed to reset message visibility:", reply.error);
+        } else {
+          // Optionally, update the message state locally or refresh the drawer
+          selectedMessage = reply.updated_message;
+          refreshLogs();
+        }
+      },
+    );
+  }
+
+  function acknowledgeMessage(messageId) {
+    live.pushEvent(
       "acknowledge_message",
       { message_id: messageId },
       (reply) => {
@@ -573,9 +592,8 @@
                         variant="outline"
                         size="sm"
                         on:click={() => {
-                          resetMessageVisibility(selectedMessage.id);
+                          acknowledgeMessage(selectedMessage.id);
                         }}
-                        disabled={resettingMessageVisibility}
                         class="flex items-center space-x-2"
                       >
                         <RotateCw class="h-4 w-4" />
--- a/assets/svelte/consumers/ShowMessages.svelte
+++ b/assets/svelte/consumers/ShowMessages.svelte
@@ -255,18 +255,12 @@
 
   function resetMessageVisibility(messageId) {
     live.pushEvent(
-      "reset_message_visibility",
+      "acknowledge_message",
       { message_id: messageId },
       (reply) => {
-        resettingMessageVisibility = false;
-
         if (reply.error) {
           // Handle error (e.g., show a toast notification)
-          console.error("Failed to reset message visibility:", reply.error);
-        } else {
-          // Optionally, update the message state locally or refresh the drawer
-          selectedMessage = reply.updated_message;
-          refreshLogs();
+          console.error("Failed to acknowledge:", reply.error);
         }
       },
     );
@@ -573,6 +567,22 @@
                       </div>
                     </div>
                   {/if}
+                  <div>
+                    <div class="bg-gray-50 p-4 rounded-lg space-y-4">
+                      <Button
+                        variant="outline"
+                        size="sm"
+                        on:click={() => {
+                          resetMessageVisibility(selectedMessage.id);
+                        }}
+                        disabled={resettingMessageVisibility}
+                        class="flex items-center space-x-2"
+                      >
+                        <RotateCw class="h-4 w-4" />
+                        <span>Acknowledge</span>
+                      </Button>
+                    </div>
+                  </div>
 
                   <!-- Timestamps Section -->
                   <div>
--- a/assets/svelte/consumers/ShowMessages.svelte
+++ b/assets/svelte/consumers/ShowMessages.svelte
@@ -21,10 +21,12 @@
   export let live;
   export let paused: boolean = false;
   export let showAcked: boolean = true;
+  export let consumerType = null;
 
   let page: number = 0;
   let loading = false;
   let selectedMessage = null;
+  let alreadyAcked = false;
   let isDrawerOpen = false;
   let messageData = null;
   let messageDataError = null;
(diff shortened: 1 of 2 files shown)


lib/sequin_web/live/sink_consumers/show.ex
```diff
@@ -302,6 +302,14 @@ defmodule SequinWeb.SinkConsumersLive.Show do
     end
   end
 
+  def handle_event("acknowledge_message", %{"message_id" => message_id}, socket) do
+    message = Enum.find(socket.assigns.messages, &(&1.id == message_id))
+    consumer = socket.assigns.consumer
+    Consumers.ack_messages(consumer, [message.ack_id])
+
+    {:noreply, put_flash(socket, :toast, %{kind: :info, title: "Message acked successfully"})}
+  end
```
RTLS, line 311: @girishramnani - I would check that `message` is present in `socket.assigns.messages` here. It is possible for the server to update the assigns, removing one or more messages, and for the message to no longer be present. If that is the case, `message.ack_id` will raise.

It might make sense to have the frontend pass `ack_id` instead of `message_id` so you can avoid this race condition.

RTLS (approved):

assets/svelte/consumers/ShowMessages.svelte
```diff
@@ -573,6 +586,21 @@
...
+                        on:click={() => {
+                          acknowledgeMessage(selectedMessage.id);
+                        }}
+                        class="flex items-center space-x-2"
+                      >
+                        <RotateCw class="h-4 w-4" />
+                        <span>Acknowledge</span>
+                      </Button>
+                    </div>
+                  </div>
```
RTLS, line 603: A few notes here:
* It would be good to hide or disable this button if the message is acknowledged (`selectedMessage.state === "delivered"`)
* Add a header and a description for what this does, perhaps like:
Screenshot 2025-01-02 at 10 20 55 AM

RTLS: The only issue with `selectedMessage.state === "delivered"` is that we use "delivered in two scenarios:
* For sequin stream sinks, this means the message was delivered but not necessarily acknowledged
* For all other sink types, this means the message was delivered *and* acknowledged

So you could refactor how state is encoded to the frontend (see https://github.com/sequinstream/sequin/blob/3ea01f0d254c631acda3875e61fe09890c73a644/lib/sequin_web/live/sink_consumers/show.ex#L892)

lib/sequin_web/live/sink_consumers/show.ex
girishramnani: So I should do both? Send the ack_id and also check if the message is present in the messages?

assets/svelte/consumers/ShowMessages.svelte
girishramnani: Can u elaborate?

RTLS: Yes- in `lib/sequin_web/live/sink_consumers/show.ex` we encode the state of messages before passing them to the frontend like this:

```
  defp get_message_state(_consumer, %AcknowledgedMessage{}), do: "delivered"
  defp get_message_state(_consumer, %{deliver_count: 0}), do: "not delivered"

  defp get_message_state(consumer, %{not_visible_until: not_visible_until, state: state}) do
    cond do
      state == :delivered and consumer.type == :sequin_stream ->
        "delivered"

      state == :delivered ->
        "delivering"

      DateTime.after?(not_visible_until, DateTime.utc_now()) ->
        "backing off"

      true ->
        "pending re-delivery"
    end
  end
```

As you can see, we use `"delivered"` in two scenarios:
1) when we have an `AcknowledgedMessage` and
2) when `state == :delivered and consumer.type == :sequin_stream`

The reason we did #2 is because sequin_stream sinks are the only sink where Sequin the application does not control acknowledgement. Instead, the user has to use the API to acknowledge a message.

For the purpose of your PR I would recommend including a new state `acknowledged` which is used exclusively for `consumer.type == :sequin_stream` and when we have an `AcknowledgedMessage`, so like:

```
  defp get_message_state(%{type: :sequin_stream}, %AcknowledgedMessage{}), do: "acknowledged"
  defp get_message_state(%{type: _}, %AcknowledgedMessage{}), do: "delivered"
```

Then I would include a check on your new acknowledged button to disable it if the `message.state == "acknowledged"` or if `message.state == "delivered" and consumer.type != "sequin_stream"`.

RTLS: When you add the `acknowledged` state, be sure to check for places this file and the ShowMessages.svelte file for `delivered` and possible include `acknowledged`. For instance, when we assign colors:

```
diff --git a/lib/sequin_web/live/sink_consumers/show.ex b/lib/sequin_web/live/sink_consumers/show.ex
index 91768511..effe5b28 100644
--- a/lib/sequin_web/live/sink_consumers/show.ex
+++ b/lib/sequin_web/live/sink_consumers/show.ex
@@ -838,7 +838,7 @@ defmodule SequinWeb.SinkConsumersLive.Show do
           data: nil,
           trace_id: message.trace_id,
           state: state,
-          state_color: "green"
+          state_color: get_message_state_color(consumer, state)
         }
     end
   end
@@ -889,6 +889,7 @@ defmodule SequinWeb.SinkConsumersLive.Show do
     end
   end

+  defp get_message_state(%{type: :sequin_stream}, %AcknowledgedMessage{}), do: "acknowledged"
   defp get_message_state(_consumer, %AcknowledgedMessage{}), do: "delivered"
   defp get_message_state(_consumer, %{deliver_count: 0}), do: "not delivered"

@@ -910,6 +911,7 @@ defmodule SequinWeb.SinkConsumersLive.Show do

   defp get_message_state_color(%{type: :sequin_stream}, state) do
     case state do
+      "acknowledged" -> "green"
       "delivered" -> "blue"
       "backing off" -> "yellow"
       _ -> "gray"
```

lib/sequin_web/live/sink_consumers/show.ex
RTLS: either one- if you pass `ack_id` from the frontend it is safe to use that without further checks. If you use `message_id` then you have to check the socket assigns.

assets/svelte/consumers/ShowMessages.svelte
girishramnani: addressed

girishramnani: please check @RTLS

lib/sequin_web/live/sink_consumers/show.ex
girishramnani: done

girishramnani:
```
Reported success to GitHub
Signed off on d24d92f64564d82291718270314b36cf39e28eab in 11 seconds
```

girishramnani:
not sure why its not signing off

lib/sequin_web/live/sink_consumers/show.ex
```diff
@@ -302,10 +303,9 @@ defmodule SequinWeb.SinkConsumersLive.Show do
...
   end
 
-  def handle_event("acknowledge_message", %{"message_id" => message_id}, socket) do
-    message = Enum.find(socket.assigns.messages, &(&1.id == message_id))
+  def handle_event("acknowledge_message", %{"ack_id" => ack_id}, socket) do
     consumer = socket.assigns.consumer
-    Consumers.ack_messages(consumer, [message.ack_id])
+    Consumers.ack_messages(consumer, [ack_id])
 
     {:noreply, put_flash(socket, :toast, %{kind: :info, title: "Message acked successfully"})}
```
RTLS, line 310:
```
{:reply, %{ok: true, updated_message: encode_message(consumer, updated_message)}, load_consumer_messages(socket)}
```

assets/svelte/consumers/ShowMessages.svelte
```diff
@@ -266,23 +272,20 @@
...
-          console.error("Failed to acknowledge:", reply.error);
-        }
-      },
-    );
+  function acknowledgeMessage(ackId) {
+    live.pushEvent("acknowledge_message", { ack_id: ackId }, (reply) => {
+      if (reply.error) {
+        // Handle error (e.g., show a toast notification)
+        console.error("Failed to acknowledge:", reply.error);
+      }
```
RTLS, line 287: `if reply.ok { selectedMessage = reply.updated_message}`

assets/svelte/consumers/ShowMessages.svelte
```diff
@@ -587,14 +590,21 @@
...
                         variant="outline"
                         size="sm"
                         on:click={() => {
-                          acknowledgeMessage(selectedMessage.id);
+                          acknowledgeMessage(selectedMessage.ack_id);
+                          alreadyAcked = true;
                         }}
                         class="flex items-center space-x-2"
+                        disabled={alreadyAcked}
                       >
```
RTLS, line 608: Let's have two states:
1. if not acked, this looks great
2. if already acked, let's `hide` the button and change the `<p>` text to `This message has been acknowledged.`

RTLS (approved):
This looks great! Approved ✅ 

Please just make two changes before merging:
1. Instead of using a toast, let's return the updated message to the frontend and use that to update the selected message state
2. Update the FE to make it clear when a message has already been acknowledged

I think this will make the user experience really great. When they press acknowledge, the left-hand sidebar will update in place to show that the message was just acknowledged :)

assets/svelte/consumers/ShowMessages.svelte
girishramnani: done

assets/svelte/consumers/ShowMessages.svelte
girishramnani: done

lib/sequin_web/live/sink_consumers/show.ex
girishramnani: done

girishramnani:
```
Reported success to GitHub
Signed off on 25da867abf147161534a2b39fa71b1344e7669b9 in 13 seconds
```

PR #798, opened 2025-01-02
Sort tables by schema and name on "Create * Sink" page

abc3:
This small change improves the UI by sorting the `filteredTables` variable by schema and name. The update ensures a more organized and predictable table display for better user experience.

before:

after:

### DIFF
--- a/assets/svelte/components/TableSelector.svelte
+++ b/assets/svelte/components/TableSelector.svelte
@@ -74,12 +74,17 @@
     if (selectedDatabaseId) {
       selectedDatabase = databases.find((db) => db.id === selectedDatabaseId);
     }
-    filteredTables =
+    filteredTables = (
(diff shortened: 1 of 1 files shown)


RTLS (commented):
Thanks @abc3 this is a good improvement!

We currently sort the tables here: https://github.com/sequinstream/sequin/blob/3ea01f0d254c631acda3875e61fe09890c73a644/lib/sequin_web/live/components/consumer_form.ex#L727

You can see that we also want tables with non-nil `sort_column` to appear first- these are tables that have been used in other sinks before. I would suggest either moving your sort logic to `consumer_form.ex` (which would just require including `schema`) or incorporating `sort_column` in your sorting here

abc3:
ahh, missed your comment :confounded:

PR #800, opened 2025-01-03
✨ Improve logs with metadata

acco:
Before, we weren't including metadata in log output (in dev or self-hosted prod!) This commit includes all metadata in our log output. We display metadata as `key=value` after the message.

To do so, we're particular about what keys we display. In dev, we don't need keys like `:domain` or `:application` - just noise. But in prod, it's useful to include those, as users will likely use to discriminate Sequin logs from other logs piping into their system.

For a list of the default metadata keys, see the comment in `lib/console_logger.ex`.

### DIFF
--- a/config/prod.exs
+++ b/config/prod.exs
@@ -9,8 +9,7 @@ import Config
 self_hosted = System.get_env("SELF_HOSTED", "0") in ~w(1 true)
 
 if self_hosted do
-  config :logger,
-    level: :info
+  config :sequin, Sequin.ConsoleLogger, drop_metadata_keys: [:mfa]
 else
   config :logger,
     level: :info,
--- a/config/config.exs
+++ b/config/config.exs
@@ -14,10 +14,10 @@ config :esbuild, :version, "0.17.11"
 
 # Configures Elixir's Logger
 config :logger, :console,
-  format: "$time $metadata[$level] $message\n",
-  metadata: [:request_id]
+  level: :info,
+  format: {Sequin.ConsoleLogger, :format},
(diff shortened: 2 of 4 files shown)


acco:
* **#800** <a href="https://app.graphite.dev/github/pr/sequinstream/sequin/800?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/sequinstream/sequin/800?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

yordis (commented):
Normally, you could reserve a key—let's call it `context` — where you can insert any information you want into the logger. This information will be included in the final output, provided that it can be marshalled into JSON.

Unless you go with the opposite side of controlling absolutely everything, how are you thinking about?

config/prod.exs
```diff
@@ -9,8 +9,7 @@ import Config
 self_hosted = System.get_env("SELF_HOSTED", "0") in ~w(1 true)
 
 if self_hosted do
-  config :logger,
-    level: :info
+  config :sequin, Sequin.ConsoleLogger, drop_metadata_keys: [:line, :file, :mfa]
```
RTLS, line 12: line and file seem useful to the interested developer. we get these under the `logger` key in prod but that migh tbe the `LoggerJSON.Formatters.Datadog` from below? In any case, I think we should include file/line in logger data somewhere in self-hosted

RTLS (approved):

PR #801, opened 2025-01-03
:sparkles: Enable crash reporting on self-hosted

acco:
SENTRY_DSN is not a secret, hence why we include in the release. We're not including in the commit/code just to make it slightly harder to abuse (make it hard for a bot to scrape).

Sentry's recommendation is to make the DSN something we can swap out without having customers pull a new release. We can do this in future with an API endpoint on our side that you can call to pull the latest DSN. And just pull that DSN on boot.


acco:
* **#801** <a href="https://app.graphite.dev/github/pr/sequinstream/sequin/801?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/sequinstream/sequin/801?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

.github/workflows/docker-build.yml
```diff
@@ -67,6 +67,8 @@ jobs:
           build-args: |
             SELF_HOSTED=1
             RELEASE_VERSION=${{ inputs.version }}
+            CURRENT_GIT_SHA=${{ github.sha }}
+            SENTRY_DSN=${{ secrets.SELF_HOSTED_SENTRY_DSN }}
```
acco, line 71: As you'll see, we move to a compile-time variable vs a runtime one (that way, you don't need to boot the docker image with the DSN env var)

.settings.example.json
```diff
@@ -1,4 +1,5 @@
 {
   "homebrewDir": "/Users/acco/x/homebrew-sequin",
-  "infraDir": "/Users/acco/x/infra"
+  "infraDir": "/Users/acco/x/infra",
+  "sentryDSN": "https://todo-token@o398678.ingest.us.sentry.io/todo-id"
```
acco, line 4: Again, not a secret - but good practice to not let the bots get their hands on it easily?

Dockerfile
```diff
@@ -66,6 +66,9 @@ WORKDIR /app
 # compile assets
 RUN mix assets.deploy
 
+# Ensure stacktraces we send to Sentry are complete
+RUN mix sentry.package_source_code
```
acco, line 70: We were missing this - possibly huge improvement!

config/prod.exs
```diff
@@ -17,6 +17,10 @@ else
     default_handler: [formatter: {LoggerJSON.Formatters.Datadog, metadata: :all}]
 end
 
+config :sentry,
+  dsn: System.get_env("SENTRY_DSN"),
+  release: System.get_env("CURRENT_GIT_SHA")
```
acco, line 22: We don't really want to do `fetch_env!`, as that breaks `MIX_ENV=prod mix compile`

So, we validate presence on boot later

lib/mix/tasks/buildpush.ex
```diff
@@ -51,7 +51,8 @@ defmodule Mix.Tasks.Buildpush do
             ecr_url: "{accnt-id}.dkr.ecr.us-east-1.amazonaws.com",
             image_name: "sequin",
             slack_bots_webhook_url: "someurl",
-            buddy_webhook_token_main: "secret"
+            buddy_webhook_token_main: "secret",
+            sentry_dsn: "https://sentry.io/some-dsn"
```
acco, line 55: Also need for our cloud build script

RTLS (approved):

PR #806, opened 2025-01-03
🏗️ SlotMessageStore lifecycle

RTLS:


RTLS:
* **#806** <a href="https://app.graphite.dev/github/pr/sequinstream/sequin/806?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/sequinstream/sequin/806?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

lib/sequin/databases_runtime/message_handler.ex
```diff
@@ -106,28 +109,36 @@ defmodule Sequin.DatabasesRuntime.SlotProcessor.MessageHandler do
...
+        # key function
+        fn {{_action, _event_or_record}, consumer} -> consumer end,
+        # value function
+        fn {{_action, event_or_record}, _consumer} -> event_or_record end
+      )
 
     res =
       Repo.transact(fn ->
-        with {:ok, count} <- insert_or_delete_consumer_messages(messages),
+        with {:ok, count} <- call_consumer_message_stores(messages_by_consumer),
```
acco, line 123: the count propagation threw me off for a second -- if I'm not mistaken, we can just use `length(messages)`? Then `call_consumer_message_stores ` can just do an `Enum.each` and return `:ok`

lib/sequin/databases_runtime/slot_message_store.ex
```diff
@@ -47,77 +49,84 @@ defmodule Sequin.ConsumersRuntime.ConsumerMessageStore do
...
       field :flush_batch_size, non_neg_integer()
       field :test_pid, pid() | nil
+      field :skip_backfill?, boolean(), default: false
     end
 
     def put_messages(%State{} = state, messages) when event_messages?(state) do
       %{state | messages: Map.merge(state.messages, messages)}
     end
 
     def put_messages(%State{} = state, messages) when record_messages?(state) do
```
acco, line 59: This is all good and I know I'm late -- but reflecting on this, what makes this so complicated is that we're collapsing multiple updates for a single record into one event. (And this collapse will only be enjoyed/realized if we're not able to move data to the sink fast enough).

I wonder if we should drop this requirement to make this simpler? Am I forgetting something beneficial about this requirement?

It's a cool thing, but not the high bit of `record` messages?

lib/sequin/databases_runtime/slot_message_store.ex
```diff
@@ -238,13 +248,36 @@ defmodule Sequin.ConsumersRuntime.ConsumerMessageStore do
     GenServer.call(via_tuple(consumer_id), {:nack, ack_ids_with_not_visible_until})
   end
 
+  @doc """
+  Peak at SlotMessageStore state. For use in dev/test only.
```
acco, line 252: 🫣 
```suggestion
  Peek at SlotMessageStore state. For use in dev/test only.
```

lib/sequin/databases_runtime/slot_message_store.ex
```diff
@@ -260,109 +293,177 @@ defmodule Sequin.ConsumersRuntime.ConsumerMessageStore do
...
-      |> Stream.map(fn msg -> %{msg | ack_id: UUID.uuid4(), dirty: true} end)
-      |> Map.new(&{&1.record_pks, &1})
+    {time, state} =
+      :timer.tc(fn ->
+        messages =
+          messages
+          # This could be dangerous in prod but is useful in test 🤔
+          |> Stream.map(fn
+            %{ack_id: ack_id} = msg when is_binary(ack_id) -> %{msg | dirty: true}
+            msg -> %{msg | ack_id: UUID.uuid4(), dirty: true}
```
acco, line 311: Got it - is idea here that we're adding a UUID if the caller didn't add one?

Yordis intro'd a good pattern for `DateTime.utc_now/0`, `Sequin.utc_now/0`. Idea is that these indeterminate functions should be easily mockable in our test suite, so we can control the state of the world.

I just added `Sequin.uuid4/0` to main:

https://github.com/sequinstream/sequin/pull/822

The only thing to keep in mind with all this is that, inside a GenServer's init, we'll need to do a `Mox.allow` to associate the GenServer with the appropriate test_pid

(You can do Mox.allow from outside a GenServer after start_link, but that feels racy!)

You can see where we do this with DateTime. The explicit allowances are OK, but I wonder if we'll want a generic pattern where if you're a GenServer and you're booted with a test_pid, you just get all the allowances for mocks associated with that test_pid. 

For an example of another place we need to do this, we'll need to do the same thing with ConnectionCache, if we have one per test process

Check this out - Mox uses NimbleOwnership to track the tree of ownership:

```elixir
{:global, Mox.Server}
|> NimbleOwnership.get_owned(self()) #=> %{
  Sequin.TestSupport.DateTimeMock => %{
    {Sequin.TestSupport.DateTimeMock, :utc_now, 0} => {0, [],
     #Function<0.56664024/0 in Sequin.Case.default_stubs/1>}
  },
  Sequin.TestSupport.UUIDMock => %{
    {Sequin.TestSupport.UUIDMock, :uuid4, 0} => {0, [],
     #Function<1.56664024/0 in Sequin.Case.default_stubs/1>}
  },
  Sequin.Sinks.RedisMock => %{
    {Sequin.Sinks.RedisMock, :send_messages, 2} => {1,
     [#Function<6.116689481/2 in Sequin.ConsumersRuntime.RedisPipelineTest."test events are sent to Redis failed Redis requests result in failed events"/1>],
     nil}
  }
}
```

So if a GenServer is given a test_pid, it could do one call to link itself to all the mocks...

One way to think about this: if Mox had an allow_all, I think we'd use that over explicit allowances for each function call?

lib/sequin/databases_runtime/slot_message_store.ex
```diff
@@ -260,109 +293,177 @@ defmodule Sequin.ConsumersRuntime.ConsumerMessageStore do
...
+          end)
+          |> Map.new(&{&1.ack_id, &1})
+
+        _state = State.put_messages(state, messages)
+      end)
 
-    state = State.put_messages(state, messages)
+    Logger.info(
+      "[SlotMessageStore] Put #{length(messages)} messages in #{div(time, 1000)}ms (message_count=#{map_size(state.messages)})"
+    )
```
acco, line 320: Again this might become debug or cloud-only

lib/sequin/databases_runtime/slot_message_store.ex
```diff
@@ -260,109 +293,177 @@ defmodule Sequin.ConsumersRuntime.ConsumerMessageStore do
...
+        {_state, _messages} = State.deliver_messages(state, deliverable_messages)
+      end)
 
-    if state.consumer.message_kind == :record do
-      with {:ok, messages} <- Consumers.put_source_data(state.consumer, messages) do
-        # Split messages into valid and nil records
-        {valid_messages, nil_messages} = Enum.split_with(messages, &(&1.data.record != nil))
+    Logger.info(
+      "[SlotMessageStore] Produced #{length(messages)} messages in #{div(time, 1000)}ms (message_count=#{map_size(state.messages)})"
+    )
```
acco, line 337: ditto debug

lib/sequin/databases_runtime/slot_message_store.ex
```diff
@@ -260,109 +293,177 @@ defmodule Sequin.ConsumersRuntime.ConsumerMessageStore do
...
+      Logger.info("[SlotMessageStore] No more messages to flush")
       schedule_flush(state)
     end
 
     {:noreply, state}
   end
 
+  defp backfill(%State{skip_backfill?: true} = state), do: {:ok, state}
+
+  defp backfill(%State{} = state) do
```
acco, line 422: Name threw me off for a moment - what about something with "load" or "load from Postgres" or even more generic like "boot" or something?

lib/sequin/databases_runtime/slot_message_store.ex
acco: What do you think about adopting this pattern, now that we're always displaying metadata in the logs?

```suggestion
    Logger.info(
      "[SlotMessageStore] Put messages", count: map_size(state.messages), time_ms: div(time, 1000)"
    )
```

Should be friendly in dev and easy for machines to parse

Only downside I can think of is that metadata is not displayed in single-line in Datadog by default. But there's probably a way to navigate this...

lib/sequin/databases_runtime/slot_message_store.ex
RTLS: whoops 🏔️

lib/sequin/consumers_runtime/consumer_idempotency.ex
```diff
@@ -47,8 +49,22 @@ defmodule Sequin.ConsumersRuntime.ConsumerIdempotency do
...
+
+    with {:ok, initial_size} <- Redis.command(["ZCARD", key]),
+         {:ok, trimmed} <- Redis.command(["ZREMRANGEBYSCORE", key, "-inf", seq]),
+         {:ok, final_size} <- Redis.command(["ZCARD", key]) do
+      Logger.info("[ConsumerIdempotency] Trimmed set",
+        consumer_id: consumer_id,
+        initial_size: initial_size,
+        records_removed: trimmed,
+        final_size: final_size
+      )
```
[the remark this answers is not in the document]
RTLS: my thoughts on logging in this stack is that I can add as many logs as possible to help debug in dev as I go and we can decrease verboseness just before or even after deploying to cloud

lib/sequin/databases_runtime/message_handler.ex
RTLS: agree!

lib/sequin/databases_runtime/slot_message_store.ex
RTLS: went with load_from_postgres - we load messages and the consumer from pg

PR #807, opened 2025-01-06
feat: add jwt to nats sink

yordis:
Screenshot 2025-01-05 at 10 08 19 PM

### DIFF
--- a/lib/sequin/sinks/nats/client.ex
+++ b/lib/sequin/sinks/nats/client.ex
@@ -27,6 +27,8 @@ defmodule Sequin.Sinks.Nats.Client do
     end
   end
 
+  @test_timeout 5_000
+
   @impl Nats
   def test_connection(%NatsSink{} = sink) do
     with :ok <-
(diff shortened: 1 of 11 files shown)


lib/sequin/sinks/nats/client.ex
```diff
@@ -42,12 +44,29 @@ defmodule Sequin.Sinks.Nats.Client do
...
         end
       else
-        _ -> {:error, Error.service(service: :nats, message: "Failed to verify NATS connection")}
+        _ ->
+          {:error,
+           Error.service(
+             service: :nats,
+             message:
+               ~s(Failed to verify NATS connection: failed to send "#{payload}" to "#{subject}" subject. Verify Nats permissions)
+           )}
```
RTLS, line 69: Should we use the error from Gnat to propagate up something useful to the user?

Also, be sure to capitalize NATS everywhere

lib/sequin/sinks/nats/client.ex
```diff
@@ -42,12 +44,29 @@ defmodule Sequin.Sinks.Nats.Client do
         receive do
           {:msg, %{topic: ^subject, body: ^payload}} ->
             :ok
+
+          _ ->
+            {:error,
+             Error.service(
+               service: :nats,
+               message: ~s(Failed to verify NATS connection: did not receive "#{payload}" on "#{subject}" subject. Verify Nats permissions)
+             )}
```
RTLS, line 53: Is there going to be a useful error here? Is it possible to hit this clause?

Ditto NATS not Nats

lib/sequin/sinks/nats/client.ex
```diff
@@ -42,12 +44,29 @@ defmodule Sequin.Sinks.Nats.Client do
...
         after
-          5000 ->
-            {:error, Error.service(service: :nats, message: "Failed to verify NATS connection")}
+          @test_timeout ->
+            {:error,
+             Error.service(
+               service: :nats,
+               message: "Failed to verify NATS connection: timeout after #{@test_timeout}ms. Verify Nats permissions"
```
RTLS, line 59:
```suggestion
               message: "Failed to verify NATS connection: did not receive test ping response after #{@test_timeout}ms. Verify NATS permissions are properly setup"
```

lib/sequin/sinks/nats/client.ex
RTLS: FWIW: These test connection flows are **very** important and useful and so worth spending some time to get right / propagate as much helpful information as we can

RTLS (approved):

PR #817, opened 2025-01-07
✨ load unchanged_toasts from old

RTLS:
This works when replica identity is set to full

### DIFF
--- a/assets/svelte/consumers/ShowSink.svelte
+++ b/assets/svelte/consumers/ShowSink.svelte
@@ -1,11 +1,6 @@
 <script lang="ts">
-  import {
-    CheckCircle2,
-    ArrowUpRight,
-    HelpCircle,
-    XCircle,
-    RefreshCw,
-  } from "lucide-svelte";
+  import { CheckCircle2, HelpCircle, XCircle, RefreshCw } from "lucide-svelte";
+  import { Button } from "$lib/components/ui/button";
   import { Card, CardContent } from "$lib/components/ui/card";
   import * as Tooltip from "$lib/components/ui/tooltip";
   import { formatNumberWithCommas } from "../utils";
@@ -35,7 +30,6 @@
   import { onMount } from "svelte";
   import * as Alert from "$lib/components/ui/alert";
   import CodeWithCopy from "$lib/components/CodeWithCopy.svelte";
-  import { Button } from "$lib/components/ui/button";
 
   export let live;
   export let parent;
@@ -108,6 +102,7 @@
   let updateChart;
   let resizeObserver;
   let refreshReplicaWarningLoading = false;
+  let dismissToastWarningLoading = false;
 
   onMount(() => {
     if (metrics.messages_processed_throughput_timeseries.length > 0) {
@@ -236,6 +231,13 @@
         .attr("d", line);
     };
   }
+
+  function handleDismissToastWarning() {
+    dismissToastWarningLoading = true;
+    live.pushEventTo(`#${parent}`, "dismiss_toast_warning", {}, () => {
+      dismissToastWarningLoading = false;
+    });
+  }
 </script>
 
 <div class="flex flex-col flex-1">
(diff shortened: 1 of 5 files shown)


RTLS:
* **#817** <a href="https://app.graphite.dev/github/pr/sequinstream/sequin/817?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/sequinstream/sequin/817?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

assets/svelte/consumers/ShowSink.svelte
```diff
@@ -289,6 +305,40 @@
...
+              Dismiss
+            </Button>
+          </div>
+        </Alert.Title>
+        <Alert.Description>
+          <p class="mb-2">
+            Some record values are being delivered as "unchanged_toast" due to
+            their large size. This means these values will appear as the string
+            literal "unchanged_toast" in update messages when that column value
+            is unchanged.
```
acco, line 321: Maybe non-passive:

> Sequin is delivering some record values as the string `unchanged_toast` due to their large size.

Understood we might want to point out that this is happening because it's what the replication slot is delivering - but that distinction might not matter?

lib/sequin/databases_runtime/message_handler.ex
```diff
@@ -422,4 +429,60 @@ defmodule Sequin.DatabasesRuntime.SlotProcessor.MessageHandler do
...
+    else
+      {ctx, messages}
+    end
+  end
+
+  defp has_unchanged_toast?(%Message{fields: nil}), do: false
+
+  defp has_unchanged_toast?(%Message{fields: fields}) do
+    Enum.any?(fields, fn %Message.Field{value: value} -> value == :unchanged_toast end)
+  end
```
acco, line 446: I assume this is just an optimization short-circuit?

lib/sequin/databases_runtime/message_handler.ex
```diff
@@ -422,4 +429,60 @@ defmodule Sequin.DatabasesRuntime.SlotProcessor.MessageHandler do
...
+    Enum.reduce(ctx.consumers, ctx, fn %SinkConsumer{annotations: annotations} = consumer, ctx ->
+      if Consumers.matches_message?(consumer, message) and not Map.has_key?(annotations, @annotations_key) do
+        annotations = Map.put_new(annotations, @annotations_key, false)
+        {:ok, consumer} = Consumers.update_consumer(consumer, %{annotations: annotations})
+        %{ctx | consumers: replace_consumer(ctx.consumers, consumer)}
+      else
+        ctx
+      end
+    end)
+  end
```
acco, line 483: Making sure I'm following: if we move to health events, we won't need this, as we'll just emit an event when we see it, right?

lib/sequin/databases_runtime/message_handler.ex
```diff
@@ -422,4 +429,60 @@ defmodule Sequin.DatabasesRuntime.SlotProcessor.MessageHandler do
...
+
+  defp has_unchanged_toast?(%Message{fields: fields}) do
+    Enum.any?(fields, fn %Message.Field{value: value} -> value == :unchanged_toast end)
+  end
+
+  defp load_unchanged_toasts_from_old(%Context{} = ctx, messages) do
+    Enum.reduce(messages, {ctx, []}, fn
+      %SlotProcessor.Message{fields: fields, old_fields: nil} = message, {ctx, messages_acc} ->
+        if Enum.any?(fields, &(&1.value == :unchanged_toast)) do
+          {annotate_consumers(ctx, message), [message | messages_acc]}
```
acco, line 452: Ah - so if old_fields: nil, it's either an insert or an update with replica identity default - is that right? hence why we annotate?

lib/sequin/databases_runtime/message_handler.ex
```diff
@@ -422,4 +429,60 @@ defmodule Sequin.DatabasesRuntime.SlotProcessor.MessageHandler do
...
+        else
+          {ctx, [message | messages_acc]}
+        end
+
+      %SlotProcessor.Message{fields: fields, old_fields: old_fields} = message, {ctx, messages_acc} ->
+        updated_fields =
+          Enum.map(fields, fn
+            %SlotProcessor.Message.Field{value: :unchanged_toast} = field ->
+              old_field = Sequin.Enum.find!(old_fields, &(&1.column_attnum == field.column_attnum))
+              %{field | value: old_field.value}
```
acco, line 462: nice

acco (approved):

lib/sequin/databases_runtime/message_handler.ex
RTLS: we will need some way to permanently debounce the event. perhaps? unless we are fine accumulate many of these in the case of a dismissal

assets/svelte/consumers/ShowSink.svelte
RTLS: we need hemingway in cursor

lib/sequin/databases_runtime/message_handler.ex
RTLS: this function is used in two spots. yeah initially it's a short circuit

lib/sequin/databases_runtime/message_handler.ex
RTLS: yeah I just changed this to match on action: :update too

PR #819, opened 2025-01-08
🏗️ TableReaderServer integration with SlotMessageStore

RTLS:


lib/sequin/consumers/consumer_record_data.ex
```diff
@@ -13,6 +13,7 @@ defmodule Sequin.Consumers.ConsumerRecordData do
   @derive Jason.Encoder
   embedded_schema do
     field :record, :map
+    field :action, Ecto.Enum, values: [:insert, :update, :delete, :read]
```
acco, line 16: are these the right values for a ConsumerRecord?

acco: If we _do_ put action, it'd always be `read` -- but feels like we should have an action on there?

acco: Perhaps this is to make event/record more similar for handling. So we need to attribute the record to either the SlotProcessor or TableReader

Two ideas there:
- keep :action, but just do `:upsert` + `:read` and keep out of serializer? (as you've done)
- have a separate property on ConsumerRecord/ConsumerEvent, i.e. `source`. Could be `:slot | :table` or something

lib/sequin/databases_runtime/message_handler.ex
```diff
@@ -312,13 +312,15 @@ defmodule Sequin.DatabasesRuntime.SlotProcessor.MessageHandler do
...
     }
   end
 
   defp record_data_from_message(%Message{action: :delete} = message, consumer) do
     %ConsumerRecordData{
       record: fields_to_map(message.old_fields),
+      action: :delete,
       metadata: struct(ConsumerRecordData.Metadata, metadata(message, consumer))
     }
   end
```
acco, line 326: Not sure if we even actually _do_ anything with these - we aren't supposed to deliver deleted ConsumerRecords

Before, we'd just use a delete to clear a message out of `consumer_records`

lib/sequin/databases_runtime/slot_message_store.ex
```diff
@@ -50,14 +51,21 @@ defmodule Sequin.DatabasesRuntime.SlotMessageStore do
       field :flush_batch_size, non_neg_integer()
       field :test_pid, pid() | nil
       field :skip_backfill?, boolean(), default: false
+      field :batch_id, String.t() | nil
+      field :unflushed_batch_ack_ids, [SinkConsumer.ack_id()], default: []
```
acco, line 55: would love to see ack_ids die, but that's probably a post-merge clean-up/refactor!

lib/sequin/databases_runtime/slot_message_store.ex
```diff
@@ -126,6 +134,69 @@ defmodule Sequin.DatabasesRuntime.SlotMessageStore do
       %{state | messages: updated_messages}
     end
 
+    def put_batch(%State{} = state, messages, batch_id) do
```
acco, line 137: put_read_batch? As it's not really just a batch of messages, there's special handling here

lib/sequin/databases_runtime/slot_message_store.ex
```diff
@@ -50,14 +51,21 @@ defmodule Sequin.DatabasesRuntime.SlotMessageStore do
       field :flush_batch_size, non_neg_integer()
       field :test_pid, pid() | nil
       field :skip_backfill?, boolean(), default: false
+      field :batch_id, String.t() | nil
```
acco, line 54: `read_batch_id`? Or `table_reader_batch_id`?

lib/sequin/databases_runtime/slot_message_store.ex
acco: Same here - we may want to associate batch to table_reader more explicitly?

lib/sequin/databases_runtime/slot_message_store.ex
```diff
@@ -179,12 +250,57 @@ defmodule Sequin.DatabasesRuntime.SlotMessageStore do
...
     end
 
+    defp intake_messages(messages) do
+      Enum.map(
+        messages,
+        fn
+          # This could be dangerous in prod but is useful in test 🤔
+          %{ack_id: ack_id} = msg when is_binary(ack_id) -> %{msg | dirty: true}
+          msg -> %{msg | ack_id: UUID.uuid4(), dirty: true}
+        end
```
acco, line 300: Sequin.uuid4() ?

lib/sequin/databases_runtime/slot_message_store.ex
```diff
@@ -347,6 +465,25 @@ defmodule Sequin.DatabasesRuntime.SlotMessageStore do
     {:reply, :ok, state}
   end
 
+  @impl GenServer
+  def handle_call({:put_batch, messages, batch_id}, _from, %State{} = state) do
+    {time, state} = :timer.tc(fn -> State.put_batch(state, messages, batch_id) end)
+
+    Logger.info(
+      "[SlotMessageStore] Put batch of #{length(messages)} messages in #{div(time, 1000)}ms (message_count=#{map_size(state.messages)})"
+    )
```
acco, line 474: Same log message commentary from earlier

lib/sequin/databases_runtime/slot_message_store.ex
acco: I take it `unflushed_batch_ack_ids` is a performance concern - with it, we're able to have a single list that we're removing items from, and then we can quickly check if a batch is finished. The alternative is to annotate all messages in memory with a batch_id or provenance or whatever, but then we'd have to traverse the list every time we wanted to see if a batch was finished flushing.

Tricky! Undesirable part is that we have to bookkeep in two places when adding, removing, and flushing messages. My first take is to lean simpler, and accept the full messages scan to check if a batch is done/flushed.  I can see us screwing this up easily when making changes in future (but maybe we won't be changing those parts much?). It's only a perf hit when we're backfilling.

Alternatively, to make mistakes harder to introduce, we can maybe expand on put_messages pattern, and just make sure message manipulation flows through limited generic functions that do the bookkeeping?

acco: Looks like TableReaderServer calls batch_progress, which means we won't check what's been flushed too frequently?

I guess there's a risk that SMS locks up if the query takes too long - but it _could_ pass off the check to a Task, which could then reply to TRS...

lib/sequin/databases_runtime/table_reader_server.ex
```diff
@@ -324,6 +325,63 @@ defmodule Sequin.DatabasesRuntime.TableReaderServer do
...
+        @initial_batch_progress_timeout,
+        state.batch_check_count,
+        @max_batch_progress_timeout
+      )
+
+    actions = [{:state_timeout, timeout, :check_batch_progress}]
+    {:keep_state_and_data, actions}
+  end
+
+  def handle_event(:state_timeout, :check_batch_progress, :commit_batch, %State{} = state) do
```
acco, line 340: Makes sense!

lib/sequin/databases_runtime/slot_message_store.ex
acco: Alternatively, with the `unflushed_batch_ack_ids`, SMS could know right away if a batch is done and tell table reader...

test/sequin/table_reader_server_test.exs
```diff
@@ -6,12 +6,11 @@ defmodule Sequin.DatabasesRuntime.TableReaderServerTest do
...
 
   alias Sequin.Consumers
-  alias Sequin.Consumers.ConsumerEvent
-  alias Sequin.Consumers.ConsumerRecord
   alias Sequin.Consumers.SequenceFilter
   alias Sequin.Databases
-  # Needs to be false until we figure out how to work with Ecto sandbox + characters
   alias Sequin.Databases.ConnectionCache
+  alias Sequin.DatabasesRuntime.SlotMessageStore
+  # Needs to be false until we figure out how to work with Ecto sandbox + characters
```
acco, line 13: Looks like this comment is lost, should be up at async

lib/sequin/consumers/consumer_record_data.ex
RTLS: still need to do this

PR #845, opened 2025-01-10
✏️ Fix typo: a Sinks -> a Sink

ruslandoga:
👋 

This PR improves wording on the first page shown after registration.

Screenshot 2025-01-11 at 02 54 42

The component always uses "a" so the easiest fix was to use a singular entity name.


acco:
Thanks @ruslandoga ! Really appreciate it 🙏

PR #865, opened 2025-01-13
Made few minor QOL changes to Makefile

oneness:
1. Self documenting feature
2. Fixed up docs for couple of targets
3. Made targets PHONY by default (detail in the comments)


Makefile
```diff
@@ -1,4 +1,19 @@
...
+NORMAL="\033[0m"
+
+# Any target invoked by make target does not correspond to a file, so
+# it should always be executed. This avoids: 1. Having to remember to
+# add newly created targets as PHONY.  2. Using a catch-all target
+# (%:) with a no-op (@:) to prevent failures for non-existing targets.
+.PHONY: $(MAKECMDGOALS)
+
+help: ## Prints target: [dep1 dep1 ...]  and what it does
+	@echo -e ${BOLD}
```
acco, line 15: @oneness Is this line incomplete? Perhaps you were trying to title the help output, like this?

```suggestion
	@echo ${BOLD}Available targets${NORMAL}
```

oneness: Good catch!  Seems like you are reading my mind :).  
FYI: pushed the fix

PR #884, opened 2025-01-18
Control mem usage by shutting down SlotProcessor after mem spike

acco:


lib/sequin/health/event.ex
```diff
@@ -30,7 +30,8 @@ defmodule Sequin.Health.Event do
     :replication_slot_checked,
     :replication_connected,
     :replication_message_processed,
-    :replication_heartbeat_received
+    :replication_heartbeat_received,
```
yordis, line 33: Side note: I can expand more about it another time.

`%ReplicationMemoryLimitExceeded{}` and the `data` field could be the keys of the struct.

https://x.com/yordisprieto/status/1854630167022350596

Dockerfile
```diff
@@ -112,6 +112,7 @@ WORKDIR /home/app
 COPY --from=builder --chown=app /app/_build .
 
 COPY .iex.exs .
+RUN ln -s /home/app/prod/rel/sequin/bin/sequin /usr/local/bin/sequin
```
acco, line 115: so now we can do `sequin remote` after shelling in

PR #927, opened 2025-01-29
feat: add s2 sink

yordis:
Signed-off-by: Yordis Prieto <yordis.prieto@gmail.com>


assets/svelte/sinks/s2/S2SinkCard.svelte
```diff
@@ -0,0 +1,26 @@
...
+      <Label>Token</Label>
+      <div class="text-sm">•••••••••••••</div>
+    </div>
+
+    <div class="flex flex-col gap-2">
+      <Label>Stream</Label>
+      <div class="text-sm">{consumer.sink.stream}</div>
+    </div>
+
+    <div class="flex flex-col gap-2">
```
shikhar, line 21: not sure if this is more consistent with how other stuff is structured, but basin might make sense to have before stream, since a stream lives in a basin

lib/sequin/sinks/s2/client.ex
```diff
@@ -0,0 +1,95 @@
...
+      messages =
+        Enum.map(messages, fn message ->
+          body =
+            message
+            |> to_body()
+            |> Jason.encode!()
+
+          %S2.V1alpha.AppendRecord{
+            body: body,
+            headers: [%S2.V1alpha.Header{name: "x-sequin-event-id", value: message.id}]
```
shikhar, line 25: Good idea to propagate this as a header!

lib/sequin/sinks/s2/client.ex
```diff
@@ -0,0 +1,95 @@
...
+          }
+        end)
+
+      req = %S2.V1alpha.AppendRequest{
+        # TODO: how do we deal with message deduplication?
```
shikhar, line 30: Is this for duplicate records that may be introduced due to retries? What is the pattern for other sinks?

I know SQS provides idempotence tokens which can be used for dedupe, and I guess you could use a compacted Kafka topic to _eventually_ dedupe.

S2 does not have anything equivalent to those yet, though we plan to support compaction like Kafka.

For now, I would suggest to go with at-least-once semantics and possibility of dupes.

It is also possible to very closely manage the content of the stream using `match_seq_num`, but probably too complex for an initial integration.

lib/sequin/sinks/s2/client.ex
```diff
@@ -0,0 +1,105 @@
...
+        input: %S2.V1alpha.AppendInput{
+          stream: sink.stream,
+          records: messages,
+          # TODO: Add match_seq_num and fencing_token when we support them
+          # match_seq_num: 1,
+          # fencing_token: nil
```
shikhar, line 36: These should not be necessary, essentially they can help ensure a single writer.

lib/sequin/sinks/s2/client.ex
```diff
@@ -0,0 +1,105 @@
...
+  alias S2.V1alpha.BasinService
+  alias S2.V1alpha.StreamService
+  alias Sequin.Error
+  alias Sequin.Sinks.S2.ConnectionCache
+  alias Sequin.Consumers.S2Sink
+  alias Sequin.Consumers.ConsumerEvent
+  alias Sequin.Consumers.ConsumerRecord
+
+  @spec send_messages(S2Sink.t(), [map()]) :: :ok | {:error, any()}
+  def send_messages(%S2Sink{} = sink, messages) do
```
shikhar, line 14: A batch (single `AppendInput`) is limited to up to 1000 records or 1 MiB of data, so should probably take that into account unless the higher-level batching is known to be more conservative than that

lib/sequin/sinks/s2/client.ex
```diff
@@ -0,0 +1,105 @@
...
+        input: %S2.V1alpha.AppendInput{
+          stream: sink.stream,
+          records: messages,
+          # TODO: Add match_seq_num and fencing_token when we support them
+          # match_seq_num: 1,
+          # fencing_token: nil
+        }
+      }
+
+      case StreamService.Stub.append(channel, req) do
```
shikhar, line 40: Preferably use `append_session`, which allows you to have multiple appends in-flight, and provides acknowledgments in the order they were submitted. It's a bi-directional gRPC so client-side state management does get a bit more involved...

lib/sequin/sinks/s2/client.ex
acco: @shikhar Yes! I believe this comment is about adding something like an SQS de-dupe key

Our system is very close to exactly once - we write to disk right after delivery that a message was delivered. So the chance of re-delivery is low. But we use de-dupe when available. Given how infrequently redeliveries should happen, no big deal to launch without this. i.e. I agree with:

> For now, I would suggest to go with at-least-once semantics and possibility of dupes.

shikhar: Sounds good. We'll also think about whether we can bring in something like Kafka's idempotent producer feature into S2...

yordis:
@acco based on this conversation, I wonder if we should have some metadata either for documentation sake, or proper control for:

1. Delivery guarantee and capability of the Sink
2. Batch size and limitations around the topic


Ideally, people find that info in the sink itself, except the batch size since it matters in our end.

lib/sequin/sinks/s2/client.ex
```diff
@@ -0,0 +1,99 @@
...
+      messages =
+        Enum.map(messages, fn message ->
+          body =
+            message
+            |> to_body()
+            |> Jason.encode!()
+
+          %S2.V1alpha.AppendRecord{
+            body: body,
+            headers: [%S2.V1alpha.Header{name: "x-sequin-event-id", value: message.id}]
```
yordis, line 25: Need review here to figure out good practices around headers.

lib/sequin/sinks/s2/client.ex
yordis: We wouldnt have multiple append_session in-flight right now, maybe we can perf test and increase the complexity later?

shikhar: Aha, if the caller needs an acknowledgment for each append before submitting the next one, then there isn't a point to `append_session`.

acco (commented):
Looking good, @yordis ! 

Let's get some unit tests in here, perhaps for the S2 client? Could be nice to get one integration test for the Pipeline, but it's pretty simple

lib/sequin/consumers_runtime/s2_pipeline.ex
```diff
@@ -0,0 +1,93 @@
...
+      opts
+      |> Keyword.fetch!(:consumer)
+      |> Repo.lazy_preload([:sequence, :postgres_database])
+
+    producer = Keyword.get(opts, :producer, Sequin.ConsumersRuntime.ConsumerProducer)
+
+    Broadway.start_link(__MODULE__,
+      name: via_tuple(consumer.id),
+      producer: [
+        module: {producer, [consumer: consumer, batch_size: 1]}
```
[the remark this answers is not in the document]
yordis: I think it is worth to document and follow up on it, ideally, the Sink configuration will tell the consumer such concern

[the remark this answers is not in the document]
shikhar: Hmm one-by-one with no pipelining (multiple records in-flight) and no batching would perform pretty badly... maybe still worth having batching on count, with the caveat that if the batch bytes get too large it would fail? Maybe the batch size can be capped at 100?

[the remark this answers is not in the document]
acco: I'm down for that! Worth noting that by this stage we've fanned out to multiple workers that are writing in parallel. But agreed batches of 100 could be a good start - won't be hard for us to add a size condition to batch size requests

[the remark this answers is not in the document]
yordis: What do I do here @acco? Leave it for now and come back or what? I am guessing I just change the `max_demand` AND `batch_size`? I forgot, need to reread again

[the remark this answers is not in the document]
acco: Looking at other sinks with batch_size > 1, looks like we leave max_demand at 1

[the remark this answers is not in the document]
acco: ah right: because max_demand 1 says "give me one batch at a time", and batch_size is the number of messages in the batch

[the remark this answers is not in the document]
shikhar: Oh, good to know there is internal parallelism before this point! Curious how you partition between workers

acco:
Closing in favor of REST API!

PR #998, opened 2025-02-16
Add verbiage in Contributions page for needed pguser 

rexposadas:
This PR addresses an issue first-time contributors of the project might come across.  One caveat was that I was not able to run the `make signoff` call because it was failing locally. 

I was going through the Contribution page. When running `mix setup`, you may encounter an error if the default PostgreSQL user "postgres" does not exist. This can be resolved by running `createuser -s postgres`. This step is only required once.

```
[error] Postgrex.Protocol (#PID<0.561.0>) failed to connect: ** (Postgrex.Error) FATAL 28000 (invalid_authorization_specification) role "postgres" does not exist
```

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -19,6 +19,12 @@ Thank you for your interest in contributing to Sequin! This document outlines th
     ```bash
     docker compose up -d
     ```
+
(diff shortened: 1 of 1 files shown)


acco:
Thanks @rexposadas !

PR #1000, opened 2025-02-17
Add vector extension to docker image so mix test can pass

rexposadas:
Before: `mix test` was failing because docker image did not have the vector extension. This PR builds a pg container with vector and uses that image. 

After: `mix test` passes. 

To test: 

docker compose down
docker compose build
docker compose up

run `mix test`


RTLS (approved):
Looks great! thanks @rexposadas

PR #1001, opened 2025-02-17
Updated quickstart docs to reflect "docker ps" output

rexposadas:
What This PR Does:
	1.	Updates the number of containers running after docker ps—now two instead of three.
	2.	Updates the PostgreSQL image version displayed.


acco:
Thank you! 🙏

PR #1012, opened 2025-02-19
🐛 [TableReaderServer] Tag watermarks with replication_id

acco:
- Support multiple replication slots on one database by tagging each watermark with a replication_id
- More assertive handling of situation where we receive a flush_batch after already having flushed the batch

### DIFF
--- a/lib/sequin/databases_runtime/message_handler.ex
+++ b/lib/sequin/databases_runtime/message_handler.ex
@@ -166,7 +166,8 @@ defmodule Sequin.DatabasesRuntime.SlotProcessor.MessageHandler do
 
   @impl MessageHandlerBehaviour
   def handle_logical_message(ctx, commit_lsn, %LogicalMessage{prefix: @low_watermark_prefix} = msg) do
-    %{"batch_id" => batch_id, "table_oid" => table_oid, "backfill_id" => backfill_id} = Jason.decode!(msg.content)
+    content = Jason.decode!(msg.content)
+    %{"batch_id" => batch_id, "table_oid" => table_oid, "backfill_id" => backfill_id} = content
 
     # Split batches into those matching the backfill_id and others
     {matching_batches, remaining_batches} = Enum.split_with(ctx.table_reader_batches, &(&1.backfill_id == backfill_id))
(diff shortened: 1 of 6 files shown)


acco:
* **#1012** <a href="https://app.graphite.dev/github/pr/sequinstream/sequin/1012?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/sequinstream/sequin/1012?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

lib/sequin/databases_runtime/message_handler.ex
```diff
@@ -177,49 +178,63 @@ defmodule Sequin.DatabasesRuntime.SlotProcessor.MessageHandler do
       )
     end
 
-    # Create new batch and prepend to remaining batches
-    %{
+    # `replication_slot_id` added later, hence why we access it differently
+    # Safe to extract from Jason.decode! in the future
+    if Map.get(content, "replication_slot_id") == ctx.replication_slot_id do
```
acco, line 183: Added this clause

lib/sequin/databases_runtime/table_reader.ex
```diff
@@ -85,12 +85,13 @@ defmodule Sequin.DatabasesRuntime.TableReader do
...
   @emit_logical_message_sql "select pg_logical_emit_message(true, $1, $2)"
-  def with_watermark(%PostgresDatabase{} = db, backfill_id, current_batch_id, table_oid, fun) do
+  def with_watermark(%PostgresDatabase{} = db, replication_slot_id, backfill_id, current_batch_id, table_oid, fun) do
     payload =
       Jason.encode!(%{
         table_oid: table_oid,
         batch_id: current_batch_id,
-        backfill_id: backfill_id
+        backfill_id: backfill_id,
+        replication_slot_id: replication_slot_id
```
acco, line 94: Add to the watermark

lib/sequin/databases_runtime/table_reader_server.ex
```diff
@@ -53,10 +53,16 @@ defmodule Sequin.DatabasesRuntime.TableReaderServer do
 
   def discard_batch(backfill_id, batch_id) when is_binary(backfill_id) do
     GenStateMachine.call(via_tuple(backfill_id), {:discard_batch, batch_id})
+  catch
+    :exit, _ ->
+      :ok
```
acco, line 58: If the TRS is not up for whatever reason, don't crash the SlotProcessor

lib/sequin/databases_runtime/table_reader_server.ex
```diff
@@ -338,13 +345,14 @@ defmodule Sequin.DatabasesRuntime.TableReaderServer do
         {:flush_batch, %{batch_id: batch_id} = batch_info},
         _state_name,
         %State{batch_id: batch_id, batch: batch} = state
-      ) do
+      )
+      when is_list(batch) do
```
acco, line 349: This bug surfaced because after flushing, I dropped the `batch` from state.

So then the second `flush_batch` call caused a downstream error (because `batch` was `nil`)

I dropped it for memory reasons, but makes sense to all do it for invariance - we already flushed the batch, so we should not allow anyone else to flush it.

lib/sequin/databases_runtime/table_reader_server.ex
```diff
@@ -338,13 +345,14 @@ defmodule Sequin.DatabasesRuntime.TableReaderServer do
...
     batch = drop_messages_by_pk(batch, batch_info.drop_pks)
     batch_size = length(batch)
 
     if batch_size == 0 do
       # Complete the batch immediately
-      state = complete_batch(state)
+      state = complete_batch(%{state | batch: nil, batch_size: 0})
```
acco, line 355: Hit a bug when we call complete_batch here, fixed by adding batch_size

lib/sequin/databases_runtime/table_reader_server.ex
```diff
@@ -365,6 +373,14 @@ defmodule Sequin.DatabasesRuntime.TableReaderServer do
     end
   end
 
+  def handle_event({:call, from}, {:flush_batch, %{batch_id: batch_id}}, _state_name, %State{
+        batch_id: batch_id,
+        batch: nil
+      }) do
+    Logger.warning("[TableReaderServer] Received flush_batch call after already flushing batch #{batch_id}. Restarting.")
+    {:stop, :normal, [{:reply, from, :ok}]}
+  end
```
acco, line 382: We should not expect to see these anymore

test/support/factory/factory.ex
```diff
@@ -31,7 +31,7 @@ defmodule Sequin.Factory do
   def http_method, do: one_of([:get, :post, :put, :patch, :delete])
   def index, do: Faker.random_between(1, 250_000)
   def integer, do: Faker.random_between(1, 100_000)
-  def non_neg_integer, do: Faker.random_between(0, 100_000)
+  def non_neg_integer, do: Faker.random_between(1, 100_000)
```
acco, line 34: postgres does not consider 0 positive

lib/sequin/databases_runtime/message_handler.ex
RTLS: should we version the logical message instead?

lib/sequin/databases_runtime/message_handler.ex
```diff
@@ -177,49 +178,63 @@ defmodule Sequin.DatabasesRuntime.SlotProcessor.MessageHandler do
...
+      case Enum.find(ctx.table_reader_batches, &(&1.batch_id == batch_id)) do
+        nil ->
+          Logger.error(
+            "[MessageHandler] Batch not found in table reader batches (batch_id: #{batch_id}, backfill_id: #{backfill_id})"
+          )
+
+          ctx.table_reader_mod.discard_batch(backfill_id, batch_id)
```
RTLS, line 218: do we need to assert success here? I see it wasn't happening before your changes but could be good

lib/sequin/databases_runtime/message_handler.ex
```diff
@@ -177,49 +178,63 @@ defmodule Sequin.DatabasesRuntime.SlotProcessor.MessageHandler do
...
+              drop_pks: batch.primary_key_values
+            })
+
+          update_in(ctx.table_reader_batches, fn batches ->
+            Enum.reject(batches, &(&1.batch_id == batch_id))
+          end)
+      end
+    else
+      # If we're sharing the same Postgres database between multiple replication slots,
+      # we may receive low watermark messages for other replication slots.
```
RTLS, line 235: wow! good call

lib/sequin/databases_runtime/table_reader_server.ex
RTLS: can be good to log a warning in this case

lib/sequin/databases_runtime/table_reader_server.ex
RTLS: good catch!

test/support/factory/factory.ex
RTLS: interesting- for which test did this matter? 0 is definitely non_negative

lib/sequin/databases_runtime/message_handler.ex
acco: @RTLS If we can't reach the TRS to tell it a batch is invalid, we should be OK: it will eventually discard when it sees we've moved on past the LSN

PR #1049, opened 2025-02-28
[stash] Call TRS to drop seen PKs

acco:

PR #1051, opened 2025-03-02
♻️ Refactor TableReaderServer pk filtering

acco:
Instead of accumulating primary keys in MessageHandler.Context, use TableReaderServer.seen_pks

To create a batch, TableReaderServer first fetches a list of PKs for its current cursor. It stores them in a public ETS table. Then makes a subsequent fetch to fetch the whole batch, writing a high watermark when it's done.

In the time before receiving the high watermark and flushing the batch, the system (SlotProcessor) will call `TableReaderServer.pks_seen`. This will drop PKs in the set from the set.

Then when the high watermark arrives, the SlotProcessor will call the TableReaderServer to flush the batch. Before flushing the batch, the batch is filtered out against the PK list. PKs that were seen during processing/change capture will be dropped from the batch, ensuring we do not have any ordering issues between real-time change messages and backfill/read messages.

### DIFF
--- a/lib/sequin/application.ex
+++ b/lib/sequin/application.ex
@@ -39,6 +39,7 @@ defmodule Sequin.Application do
   defp children(_) do
     base_children() ++
       [
+        SequinWeb.Telemetry,
         MutexedSupervisor.child_spec(
           Sequin.Runtime.MutexedSupervisor,
           [
@@ -56,7 +57,6 @@ defmodule Sequin.Application do
     Sequin.Redis.connect_cluster()
 
     [
-      SequinWeb.Telemetry,
       Sequin.Repo,
       Sequin.Vault,
       Sequin.PubSub.child_spec(),
--- a/lib/sequin/databases_runtime/message_handler.ex
+++ b/lib/sequin/databases_runtime/message_handler.ex
@@ -55,7 +55,6 @@ defmodule Sequin.DatabasesRuntime.SlotProcessor.MessageHandler do
       field :replication_slot_id, String.t()
       field :postgres_database, PostgresDatabase.t()
       field :table_reader_mod, module(), default: TableReaderServer
-      field :running_backfill_ids, MapSet.t(String.t()), default: MapSet.new()
     end
   end
 
@@ -72,15 +71,15 @@ defmodule Sequin.DatabasesRuntime.SlotProcessor.MessageHandler do
   end
 
   @impl MessageHandlerBehaviour
-  def handle_messages(%Context{} = ctx, []) do
-    {:ok, 0, ctx}
+  def handle_messages(%Context{}, []) do
+    {:ok, 0}
   end
 
   def handle_messages(%Context{} = ctx, messages) do
     execute_timed(:handle_messages, fn ->
       Logger.debug("[MessageHandler] Handling #{length(messages)} message(s)")
 
-      {ctx, messages} = load_unchanged_toasts(ctx, messages)
+      messages = load_unchanged_toasts(ctx, messages)
 
       messages_by_consumer_id =
         execute_timed(:messages_by_consumer_id, fn ->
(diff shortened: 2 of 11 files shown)


acco:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/sequinstream/sequin/1051?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1051** <a href="https://app.graphite.dev/github/pr/sequinstream/sequin/1051?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/sequinstream/sequin/1051?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1049** <a href="https://app.graphite.dev/github/pr/sequinstream/sequin/1049?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

lib/sequin/application.ex
```diff
@@ -39,6 +39,7 @@ defmodule Sequin.Application do
   defp children(_) do
     base_children() ++
       [
+        SequinWeb.Telemetry,
```
acco, line 42: move out of tests, as we now have a poller job in there

lib/sequin/databases_runtime/message_handler.ex
```diff
@@ -72,15 +71,15 @@ defmodule Sequin.DatabasesRuntime.SlotProcessor.MessageHandler do
   end
 
   @impl MessageHandlerBehaviour
-  def handle_messages(%Context{} = ctx, []) do
-    {:ok, 0, ctx}
+  def handle_messages(%Context{}, []) do
+    {:ok, 0}
```
acco, line 75: we don't return `ctx` anymore, as we're not tracking pks here

lib/sequin/databases_runtime/message_handler.ex
```diff
@@ -72,15 +71,15 @@ defmodule Sequin.DatabasesRuntime.SlotProcessor.MessageHandler do
...
   end
 
   def handle_messages(%Context{} = ctx, messages) do
     execute_timed(:handle_messages, fn ->
       Logger.debug("[MessageHandler] Handling #{length(messages)} message(s)")
 
-      {ctx, messages} = load_unchanged_toasts(ctx, messages)
+      messages = load_unchanged_toasts(ctx, messages)
```
acco, line 82: we never actually modified ctx down there?

lib/sequin/databases_runtime/slot_processor.ex
```diff
@@ -103,6 +105,8 @@ defmodule Sequin.DatabasesRuntime.SlotProcessor do
       field :accumulated_msg_binaries, %{count: non_neg_integer(), bytes: non_neg_integer(), binaries: [binary()]},
         default: %{count: 0, bytes: 0, binaries: []}
 
+      field :backfill_watermark_messages, [LogicalMessage.t()], default: []
```
acco, line 108: Before, we processed logical messages (backfill watermark messages) right away. This was actually a bug - handling logical messages right away means they're handled _in front_ of any acc messages. This means out-of-order processing.

Now, as you'll see, we accumulate backfill watermark messages, and then flush them after flushing our other acc binaries.

This is temp - goal in subsequent refactors is to make SlotProcessor do less / be dumber

lib/sequin/databases_runtime/table_reader_server.ex
```diff
@@ -15,6 +15,7 @@ defmodule Sequin.DatabasesRuntime.TableReaderServer do
   alias Sequin.Error
   alias Sequin.Error.InvariantError
   alias Sequin.Error.ServiceError
+  alias Sequin.EtsMultiset
```
acco, line 18: added this right to main

PR #1086, opened 2025-03-11
Add Docker Compose setup mention to README

abc3:
PostgreSQL and Redis containers are essential for local development, as Sequin will not function without them


acco:
Thanks as always, @abc3 !

PR #1087, opened 2025-03-11
Add time units to timer.tc calls

abc3:
The `timer.tc` function has [supported time units](https://github.com/erlang/otp/blob/05737d130706c7189a8e6750d9c2252d2cc7987e/lib/stdlib/src/timer.erl#L578) since Erlang 26. 

Because the [`.tool-versions` file](https://github.com/sequinstream/sequin/blob/f3cf9b2772aaa954d4eaa66f9377338d9c9c02e9/.tool-versions#L2) requires Erlang 27.2.1, using `timer.tc` with the `:millisecond` argument is safe and eliminates the need for unnecessary conversions.

### DIFF
--- a/lib/sequin/bench/bench.ex
+++ b/lib/sequin/bench/bench.ex
@@ -196,8 +196,8 @@ defmodule Sequin.Bench do
   end
 
   defp exec_fun(scenario_fun, input_args) do
-    {execution_time, _result} = :timer.tc(fn -> apply(scenario_fun, input_args) end)
-    execution_time / 1000
+    {execution_time, _result} = :timer.tc(fn -> apply(scenario_fun, input_args) end, :millisecond)
+    execution_time
   end
 
   defp format_stats(test_name, stats) do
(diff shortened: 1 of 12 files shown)


abc3:
I have no clue how that signoff must work. The branch in the cloned repo is [signed](https://github.com/abc3/sequin/commit/5bc2723db57244c78ff34d16cea3d807835a8f24), but in this PR, it doesn't recognize that `¯\_(ツ)_/¯`

acco (approved):

acco:
@abc3 TIL! Thanks so much, nice improvement

acco:
@abc3 Also, we'll add GH actions ... I can see how `make signoff` may just not work for an open source project

abc3:
@acco Nice! Yeah, and signoff also blocks the ability to send a proper pr made via the web ui. 
Some sources for GA inspiration: 
- https://github.com/supabase/realtime/blob/main/.github/workflows/tests.yml
- https://github.com/supabase/supavisor/blob/main/.github/workflows/elixir.yml

PR #1106, opened 2025-03-15
Move sort_column_attnum to backfills

acco:
`sequences` in general are a legacy hold-over. The last important bit on them was `sort_column_attnum`, which dictated the sort column used for backfills. But the problem is that sequences are hidden in the UI, and so the sort column can't be changed after create. Also, because sort columns are *only* related to backfills, it makes sense to tie them directly to the backfill (vs bind them to the table).

This commit moves sort_column_attnum over. It does not do clean-up of the sort_column_attnum on sequences. However, new sequences can now be created without sort_column_attnum populated. We can drop these columns in the future/at any time.

### DIFF
--- a/assets/svelte/components/BackfillForm.svelte
+++ b/assets/svelte/components/BackfillForm.svelte
@@ -1,0 +1,180 @@
+<script lang="ts">
+  import { RadioGroup, RadioGroupItem } from "$lib/components/ui/radio-group";
+  import { Label } from "$lib/components/ui/label";
+  import { Input } from "$lib/components/ui/input";
+  import {
+    Select,
+    SelectContent,
+    SelectItem,
+    SelectTrigger,
+    SelectValue,
+  } from "$lib/components/ui/select";
+  import { ExternalLinkIcon } from "lucide-svelte";
+  import type { Table } from "$lib/databases/types";
+  import Datetime from "./Datetime.svelte";
+  import { conditionalClass } from "$lib/utils";
+
+  export let form: {
+    startPosition: "beginning" | "specific";
+    sortColumnAttnum: number | null;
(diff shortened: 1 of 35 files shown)


acco:
* **#1106** <a href="https://app.graphite.dev/github/pr/sequinstream/sequin/1106?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/sequinstream/sequin/1106?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

assets/svelte/components/BackfillForm.svelte
```diff
@@ -0,0 +1,180 @@
+<script lang="ts">
```
acco, line 1: consolidate forms across consumer create and consumer show

assets/svelte/consumers/Backfill.svelte
```diff
@@ -5,11 +5,6 @@
   import { formatNumberWithCommas, formatRelativeTimestamp } from "../utils";
   import { Loader2 } from "lucide-svelte";
   import { CheckCircle2 } from "lucide-svelte";
-  import * as Dialog from "$lib/components/ui/dialog";
-  import { Input } from "$lib/components/ui/input";
-  import { RadioGroup, RadioGroupItem } from "$lib/components/ui/radio-group";
-  import { Label } from "$lib/components/ui/label";
-  import Datetime from "../components/Datetime.svelte";
```
acco, line 12: move the new backfill form out of Backfill - this is really a show component. Hoisted up modal control to ShowSinkConsumer, and then inside that modal we render BackfillForm

PR #1149, opened 2025-03-28
🗂️ Transform model (polymorphic)

RTLS:


lib/sequin/consumers/path_transform.ex
```diff
@@ -0,0 +1,31 @@
...
+    |> validate_path()
+  end
+
+  defp validate_path(changeset) do
+    changeset
+    |> validate_format(
+      :path,
+      ~r/^[a-zA-Z0-9\-._~!$&'()*+,;=:@%\/]+$/,
+      message: "must be a valid URL path"
+    )
```
acco, line 28: Incorrect validator?

Presumably the syntax is:

`record.addresses.first_line`

And they can use `""` to optionally escape parts as well, so:

`record."user_info.address".first_line`

Perhaps that's it?

lib/sequin/consumers/transform.ex
```diff
@@ -0,0 +1,80 @@
+defmodule Sequin.Transforms.Transform do
```
acco, line 1: Lives in the `consumers` folder?

If so, we should probably call it `Sequin.Consumers.Transform`?

lib/sequin/consumers/transform.ex
```diff
@@ -0,0 +1,80 @@
...
+  import PolymorphicEmbed
+
+  alias Sequin.Accounts.Account
+  alias Sequin.Consumers.PathTransform
+  alias Sequin.Databases.Sequence
+
+  @derive {Jason.Encoder, only: [:id, :name, :type, :config, :account_id, :sequence_id]}
+  schema "transforms" do
+    field :name, :string
+    field :type, :string, read_after_writes: true
```
acco, line 16: If this is for defaults, consider just setting a default up here?

Reason I prefer is that it's not clear what the read_after_writes is for otherwise

priv/repo/migrations/20250328152841_create_transforms_table.exs
```diff
@@ -0,0 +1,25 @@
+defmodule Sequin.Repo.Migrations.CreateTransformsTable do
+  use Ecto.Migration
+
+  @config_schema Application.compile_env(:sequin, [Sequin.Repo, :config_schema_prefix])
+
+  def change do
+    create table(:transforms, prefix: @config_schema) do
+      add :name, :string, null: false
+      add :type, :string, generated: "always as (config->>'type') stored"
+      add :config, :jsonb, null: false, default: "{}"
```
acco, line 10: We should remove the default, as that's not a valid config?

lib/sequin/consumers/transform.ex
acco: Ah just saw that's it's generated from the config - clever

priv/repo/migrations/20250328152841_create_transforms_table.exs
```diff
@@ -0,0 +1,25 @@
...
+      add :account_id, references(:accounts, on_delete: :delete_all, prefix: @config_schema),
+        null: false
+
+      add :sequence_id, references(:sequences, on_delete: :delete_all, prefix: @config_schema),
+        null: false
+
+      timestamps()
+    end
+
+    create unique_index(:transforms, [:account_id, :name], prefix: @config_schema)
```
acco, line 21: 👍

priv/repo/migrations/20250328152842_add_transform_to_sink_consumers.exs
```diff
@@ -0,0 +1,13 @@
+defmodule Sequin.Repo.Migrations.AddTransformToSinkConsumers do
+  use Ecto.Migration
+
+  @config_schema Application.compile_env(:sequin, [Sequin.Repo, :config_schema_prefix])
+
+  def change do
+    alter table(:sink_consumers, prefix: @config_schema) do
+      add :transform_id, references(:transforms, on_delete: :nilify_all, prefix: @config_schema)
```
acco, line 8: should we consider raising? I assume we'd want to force you to de-associate a transform with everything before removing it

acco (approved):
Looks good - see in-line comments and a question about syntax/validation

lib/sequin/consumers/path_transform.ex
acco: I see you fix!

lib/sequin/consumers/transform.ex
acco: I see you fix!

priv/repo/migrations/20250328152842_add_transform_to_sink_consumers.exs
RTLS: set to `:restrict`

priv/repo/migrations/20250328152841_create_transforms_table.exs
RTLS: 👍

PR #1160, opened 2025-03-29
✨ Trigger database update when new relation message received

acco:
When we receive a relation message for a given table for a second time, we can know ~for sure that the table was updated/migrated. So, we can trigger the DatabaseUpdateWorker to run.

This won't work if:
- we weren't connected to the database during the migration
- we were connected to the database during the migration, but we hadn't received any records pertaining to the migrated table prior to the migration (so the relation has not been cached yet)


acco:
* **#1160** <a href="https://app.graphite.dev/github/pr/sequinstream/sequin/1160?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/sequinstream/sequin/1160?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1191, opened 2025-04-03
Refresh database table schema when a relation message with updated schema arrives at stot

### LINKED ISSUE #924: Refresh database tables when we see a relation message in the replication slot
We should be using the presence of relation messages coming through the replication slot as a possible signal that schemas have changed. If schemas change, we should trigger a refresh of the `PostgresDatabase.tables`

Easiest to start: When we see an unrecognized table coming through the slot, good sign something was created.

Will help in test/dev, where tables may be getting dropped/created frequently

davoclavo:
Howdy Sequin team! Here is a proposal to fix #924

Main points:

- Add longer-lived schema hash storage in Redis using `HGET` + `HSET`
- Remove `unique_period` to avoid Oban skipping queue job due to throttling

Please let me know if there are better ways to do things, happy to adjust things.

I wanted to add automated testing for this, but I am still fighting to get my postgres+pgvector up and running. I did however run some manual tests changing table schemas and creating/updating records by hand. Any pointers in this direction of automated testing are also more than welcome.

### DIFF
--- a/lib/sequin/runtime/slot_processor_server.ex
+++ b/lib/sequin/runtime/slot_processor_server.ex
@@ -33,6 +33,7 @@ defmodule Sequin.Runtime.SlotProcessorServer do
   alias Sequin.Runtime.PostgresAdapter.Decoder.Messages.LogicalMessage
   alias Sequin.Runtime.PostgresAdapter.Decoder.Messages.Relation
   alias Sequin.Runtime.PostgresAdapter.Decoder.Messages.Update
+  alias Sequin.Runtime.PostgresRelationHashCache
   alias Sequin.Runtime.SlotMessageStore
   alias Sequin.Runtime.SlotProcessor.Message
   alias Sequin.Workers.CreateReplicationSlotWorker
@@ -952,7 +953,10 @@ defmodule Sequin.Runtime.SlotProcessorServer do
   end
 
   @spec put_relation_message(Relation.t(), State.t()) :: State.t()
-  defp put_relation_message(%Relation{id: id, columns: columns, namespace: schema, name: table}, %State{} = state) do
+  defp put_relation_message(
+         %Relation{id: id, columns: columns, namespace: schema, name: table} = relation,
+         %State{} = state
+       ) do
     conn = get_cached_conn(state)
 
     # First, determine if this is a partition and get its parent table info
@@ -1022,16 +1026,22 @@ defmodule Sequin.Runtime.SlotProcessorServer do
         end
       end)
 
-    # Check if we already have a schema for this relation ID
-    # If so, enqueue a database update worker as the schema may have changed
-    if Map.has_key?(state.schemas, id) do
-      Logger.info("[SlotProcessorServer] Received relation message for existing relation, enqueueing database update",
+    # Create a relation with enriched columns
+    enriched_relation = %{relation | columns: enriched_columns}
+
+    # Compare schema hashes to detect changes
+    current_hash = PostgresRelationHashCache.compute_schema_hash(enriched_relation)
+    stored_hash = PostgresRelationHashCache.get_schema_hash(state.postgres_database.id, id)
+
+    unless stored_hash == current_hash do
+      Logger.info("[SlotProcessorServer] Schema changes detected for table, enqueueing database update",
         relation_id: id,
         schema: parent_info.schema,
         table: parent_info.name
       )
 
-      DatabaseUpdateWorker.enqueue(state.postgres_database.id)
+      PostgresRelationHashCache.update_schema_hash(state.postgres_database.id, id, current_hash)
+      DatabaseUpdateWorker.enqueue(state.postgres_database.id, unique_period: 0)
     end
 
     # Store using the actual relation_id but with parent table info
--- a/lib/sequin/runtime/postgres_relation_hash_cache.ex
+++ b/lib/sequin/runtime/postgres_relation_hash_cache.ex
@@ -1,0 +1,80 @@
+defmodule Sequin.Runtime.PostgresRelationHashCache do
+  @moduledoc """
+  Handles computing, and caching schema hashes for PostgreSQL relations.
+
+  This module provides functionality to:
+  - Compute a hash of a relation's schema using Erlang's phash2 function
+  - Read, and update schema hashes using Redis
+
+  Schema hashes are used to track changes in relation schemas over time.
+
+  They are stored in a Redis hash using a key pattern defined in `schema_hash_key` below.
+  """
+
+  alias Sequin.Redis
+  alias Sequin.Runtime.PostgresAdapter.Decoder.Messages.Relation
+
+  require Logger
+
+  @doc """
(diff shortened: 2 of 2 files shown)


lib/sequin/runtime/schema_hash.ex
```diff
@@ -0,0 +1,80 @@
...
+
+  require Logger
+
+  @doc """
+  Computes a hash of a relation's schema.
+  """
+  @spec compute_schema_hash(Relation.t()) :: String.t()
+  def compute_schema_hash(%Relation{} = relation) do
+    # Create a string representation of the schema that includes all relevant fields
+    schema_str = Jason.encode!(relation)
```
davoclavo, line 18: Is there a more elixir idiomatic way of hashing an instance without needing json encoders?

RTLS:
```
:erlang.phash2(term)
```

This function can take any elixir term and returns a hash :)

lib/sequin/runtime/postgres_adapter/decoder.ex
```diff
@@ -84,6 +84,8 @@ defmodule Sequin.Runtime.PostgresAdapter.Decoder do
       """
       use TypedStruct
 
+      @derive Jason.Encoder
```
acco, line 87: You can remove if you use phash below!

lib/sequin/runtime/schema_hash.ex
```diff
@@ -0,0 +1,80 @@
+defmodule Sequin.Runtime.SchemaHash do
```
acco, line 1: Perhaps `PostgresRelationHashCache`?

lib/sequin/runtime/schema_hash.ex
```diff
@@ -0,0 +1,80 @@
...
+          Logger.error("[SlotProcessorServer] Error storing schema hash",
+            relation_id: relation.id,
+            error: error
+          )
+          state
+      end
+   end
+
+  @spec schema_hash_key(String.t()) :: String.t()
+  defp schema_hash_key(database_id), do: "sequin:schema_hash:#{database_id}"
```
acco, line 79:
```suggestion
  defp schema_hash_key(database_id), do: "sequin:runtime:postgres_relation_hash_cache:#{database_id}"
```

lib/sequin/runtime/schema_hash.ex
```diff
@@ -0,0 +1,80 @@
+defmodule Sequin.Runtime.SchemaHash do
+  @moduledoc """
+  Handles computing schema hash
```
acco, line 3: Expand just a little!

RTLS:
Looks great! just chatted with @acco and he is leaving comments. stoked about this PR @davoclavo :D

lib/sequin/runtime/slot_processor_server.ex
```diff
@@ -106,6 +107,7 @@ defmodule Sequin.Runtime.SlotProcessorServer do
       field :publication, String.t()
       field :replication_slot, PostgresReplicationSlot.t()
       field :schemas, %{}, default: %{}
+      field :schema_hashes, %{}, default: %{}
```
acco, line 110: I know I directed you to cache in SlotProcessorServer memory. But discussing with @RTLS , we actually think it's fine to just hit Redis every time! Simplifies/removes the synchronization problem.

lib/sequin/runtime/slot_processor_server.ex
```diff
@@ -1022,16 +1025,26 @@ defmodule Sequin.Runtime.SlotProcessorServer do
...
       )
 
-      DatabaseUpdateWorker.enqueue(state.postgres_database.id)
+      # Enqueue schema reloading without any throttling
+      DatabaseUpdateWorker.enqueue(state.postgres_database.id, unique_period: 0)
+
+      SchemaHash.update_schema_hash(state, enriched_relation, current_hash)
+    else
+      state
     end
```
acco, line 1048: Great!!

acco (commented):
@davoclavo This is really excellent, thank you!

See minor comments in-line

lib/sequin/runtime/slot_processor_server.ex
davoclavo: Happy to make that change. However, my concern is that this check gets performed on every event reported by the slot, which means Redis will be hit quite frequently.

Perhaps I'm missing something important — for example, if Sequin is running in multiple instances or in cluster mode. If it's running as a single instance, then Redis would mainly be useful for persisting the cache across restarts, right?

I’d appreciate it if you could shed some light — happy to do whatever makes the most sense long term.

cc @RTLS

davoclavo:
btw, I ran `mix format` and made some slight indentation / `if` changes. 

Is this the recommended tool to lint/format the code? I couldn't find any reference to other tools in the Makefile or mix.exs

acco:
@davoclavo Hmm... `mix format` ought to just work. You're using `asdf` yeah? That will ensure you're using the same version of elixir in `.tool-versions`. 

I assume you're saying that code you _didn't_ touch was reformatted?

lib/sequin/runtime/slot_processor_server.ex
acco: @davoclavo Thanks for asking/clarifying this!

We only receive a relation message right before a change (insert/update/delete) message IF:

- we haven't received a message from that table yet
- the table just changed

So I think hitting Redis is OK! If a db has 100 tables and all of them are going to the replication slot, it means only hitting Redis 100 times before we're done hitting it

Make sense / what do you think?

davoclavo: Makes sense, thanks for the explanation. Will make these changes so there is no in-memory cache any more, and fully rely on the Redis one

davoclavo:
> @davoclavo Hmm... `mix format` ought to just work. You're using `asdf` yeah? That will ensure you're using the same version of elixir in `.tool-versions`.
> 
> I assume you're saying that code you _didn't_ touch was reformatted?

Oh my bad for the confusion, I wanted to confirm if `mix format` was the right tool for formatting! It seems it is (I just saw it is mentioned in CONTRIBUTING.md)
No code was formatted outside of my changes, so thankfullt there were no we did not have version mismatches

Regarding `asdf` - thanks for the remainder/question. 

I am using `nix` (via [`devbox`](https://www.jetify.com/docs/devbox/)) for version management, instead of `asdf`. A useful thing about this setup is that I can specify all the code-dependencies (like how `asdf` does) - but also external dependencies, such as postgres, redis, etc and also run them as native services with `process-compose`. So in summary it is like asdf+docker.
Happy to share these config files in a separate branch in case they are useful to someone else.

davoclavo:
All suggested changes have been addressed, thanks for the feedback. Please let me know if anything else should be adjusted :)

acco:
Awesome work, @davoclavo ! Merging

acco:
FYI, added an integration test here:
https://github.com/sequinstream/sequin/pull/1196

PR #1202, opened 2025-04-06
Load logger level from LOG_LEVEL for all environments and adjust defaults

davoclavo:
Load logger level from env var `LOG_LEVEL` for all environments except test

Some notes:
- We keep the default `:info` level within [`ConfigParser.log_level/1`](https://github.com/sequinstream/sequin/blob/main/lib/sequin/config_parser.ex#L12-L18)
- We keep a hardcoded `:warn` log level for tests to reduce verbosity

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -39,6 +39,12 @@ if System.get_env("PHX_HOST") do
   Logger.warning("PHX_HOST environment variable is deprecated. Please use SERVER_HOST instead.")
 end
 
+if config_env() == :test do
+  config :logger, level: ConfigParser.log_level(env_vars, :warning)
+else
+  config :logger, level: ConfigParser.log_level(env_vars, :info)
+end
+
 # config/runtime.exs is executed for all environments, including
 # during releases. It is executed after compilation and before the
 # system starts, so it is typically used to load production configuration
@@ -119,8 +125,6 @@ if config_env() == :prod and self_hosted do
       "0" -> false
       other -> raise("Invalid SERVER_CHECK_ORIGIN: #{other}, must be true or false or 1 or 0")
     end
-
-  config :logger, level: ConfigParser.log_level(env_vars)
 
   config :sequin, Sequin.Posthog,
     req_opts: [base_url: "https://us.i.posthog.com"],
--- a/lib/sequin/config_parser.ex
+++ b/lib/sequin/config_parser.ex
@@ -9,20 +9,20 @@ defmodule Sequin.ConfigParser do
     |> put_redis_opts_ipv6(env)
   end
 
-  def log_level(env) do
+  @valid_log_levels [:error, :warning, :notice, :info, :debug]
+  def log_level(env, default_level) when default_level in @valid_log_levels do
     env
-    |> Map.get("LOG_LEVEL", "info")
+    |> Map.get("LOG_LEVEL", default_level |> to_string())
     |> String.downcase()
     |> String.to_atom()
-    |> validate_log_level()
+    |> validate_log_level(default_level)
   end
 
-  @valid_log_levels [:error, :warning, :notice, :info, :debug]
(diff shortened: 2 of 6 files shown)


config/runtime.exs
```diff
@@ -39,6 +39,8 @@ if System.get_env("PHX_HOST") do
   Logger.warning("PHX_HOST environment variable is deprecated. Please use SERVER_HOST instead.")
 end
 
+config :logger, level: ConfigParser.log_level(env_vars)
```
davoclavo, line 42: just for clarity for the reviewer: this config line was moved up from inside an `if config_env() == :prod and self_hosted do` block

acco (changes requested):
@davoclavo Thank you for this!

Only one issue: In test, the log_level is overwritten by `runtime`. So it's set to `info` and not `warning` (as desired)

How we might solve this:

In `runtime`, only use `ConfigParser` to set log level if LOG_LEVEL is set. Otherwise, don't set it -- which will fall back to whatever we set in `config`, `test`, etc. (This means you should also revert the changes in `config` and `prod`

If you do that, you might also consider making `ConfigParser.log_level` hard match on the env var being set - but I don't have strong opinions there

davoclavo:
Apologies for all the back and forth. I think I finally understood what was happening, my mental model of how config files are loaded was not correct (`runtime.exs` is the last file to be loaded).

I propose a solution of centralizing setting log levels only in `runtime.exs` and just adjust the default depending on the environment. 

So it will follow this logic:

- If `LOG_LEVEL` is set and is valid, we will use that value regardless of default for any environment
- If `LOG_LEVEL` is not set, or is invalid:
  - We default to `:warning` log level for tests to reduce verbosity
  - We default to `:info` log level for any other environment

Lmk if this proposal is complicated or non-idiomatic (e.g. having `:test` configurations in `runtime.exs`) and if I should follow your suggestion instead.

Thanks!

acco:
@davoclavo That sounds like the right solution! We ought to just completely control the log level in runtime (vs the env-specific files)

davoclavo:
Cool! These suggested changes are already pushed to this branch

config/runtime.exs
```diff
@@ -39,6 +39,12 @@ if System.get_env("PHX_HOST") do
   Logger.warning("PHX_HOST environment variable is deprecated. Please use SERVER_HOST instead.")
 end
 
+if config_env() == :test do
+  config :logger, level: ConfigParser.log_level(env_vars, :warning)
+else
+  config :logger, level: ConfigParser.log_level(env_vars, :info)
+end
```
acco, line 46: Great!

lib/sequin/config_parser.ex
```diff
@@ -9,20 +9,20 @@ defmodule Sequin.ConfigParser do
     |> put_redis_opts_ipv6(env)
   end
 
-  def log_level(env) do
+  @valid_log_levels [:error, :warning, :notice, :info, :debug]
+  def log_level(env, default_level) when default_level in @valid_log_levels do
```
acco, line 13: awesome

acco (approved):
Great work @davoclavo , thank you!

lib/sequin/config_parser.ex
```diff
@@ -9,20 +9,20 @@ defmodule Sequin.ConfigParser do
...
     env
-    |> Map.get("LOG_LEVEL", "info")
+    |> Map.get("LOG_LEVEL", default_level |> to_string())
```
acco, line 15: Formatter issue - so sorry we didn't catch this! We should have a github action that runs on opened PRs now from outside contributors that will alert about this stuff (just need to rebase) 

```suggestion
    |> Map.get("LOG_LEVEL", to_string(default_level))
```

davoclavo:
@acco awesome, thanks! 

I did forget to run `mix format` so good idea about running this check in CI for contributors as well

PR #1220, opened 2025-04-08
chore: update Elixir dependency installation to check for locked versions in PR checks workflow

yordis:


.github/workflows/pr-checks.yml
```diff
@@ -96,7 +96,7 @@ jobs:
           restore-keys: ${{ runner.os }}-build-
 
       - name: Install Elixir dependencies
-        run: mix deps.get
+        run: mix deps.get --check-locked
```
acco, line 99: Can you add to signoff.sh as well?

https://github.com/sequinstream/sequin/blob/master/scripts/signoff.sh

yordis: done

PR #1240, opened 2025-04-11
vendor Prometheus.PlugExporter

fazzone:
the [prometheus plugs library](https://github.com/prometheus-erl/prometheus-plugs) which implements the /metrics endpoint as a phoenix plug, hasn't been updated in several years.  As a result, it was giving a noisy compiler warning due to use of now-deprecated elixir syntax, as well as constraining us to use an old version of the prometheus_ex library.  It seems to be somewhat common practice to have an internal fork of the prometheus_plugs library, and we could do that too, but since we really only care about plugexporter, and we aren't super interested in the other plugs, I propose we simply vendor this one file (MIT licensed).  Phoenix itself has nice :telemetry events which could be used to monitor it in more detail which AFAICT is the currently-preferred solution for phoenix metrics.


lib/sequin/metrics/plug_exporter.ex
```diff
@@ -0,0 +1,238 @@
...
+  ##               path: "/metrics",
+  ##               format: :protobuf]]
+  defmacro __using__(_opts) do
+    module_name = __CALLER__.module
+
+    registry = Config.registry(module_name)
+    path = Plug.Router.Utils.split(Config.path(module_name))
+    auth = Config.auth(module_name)
+    format = normalize_format(Config.format(module_name))
+    Code.ensure_loaded!(:accept_header)
```
fazzone, line 78: this is the only change to the original file besides fixing the deprecated syntax warning.  the compiler complains if nothing is done.  the original library doesn't need this because it has the `accept` application as a dependency.  I am pretty confident that the [accept application ](https://github.com/prometheus-erl/accept/blob/master/src/accept.app.src) doesn't actually do anything and this is therefore fine to do

yordis:
We have been using https://hex.pm/packages/peep professionally @acco really good library, and the author is on top of it

yordis:
@RTLS I have some boilerplate code with Peep, it shouldn't take me long to help you with it.

fazzone:
if we do anything we should do #1249

PR #1249, opened 2025-04-13
chore: use peep for prometheus

yordis:


lib/sequin/metrics/prometheus.ex
```diff
@@ -254,82 +246,132 @@ defmodule Sequin.Prometheus do
...
   def set_delivery_saturation(consumer_id, consumer_name, percent) do
-    Gauge.set([name: :sequin_delivery_saturation_percent, labels: [consumer_id, consumer_name]], percent)
+    :telemetry.execute([:sequin, :delivery_saturation_percent], %{
+      percent: percent,
+      consumer_id: consumer_id,
+      consumer_name: consumer_name
+    })
   end
 
   defp validate_success(:ok), do: :ok
```
yordis, line 377: what is the idea of this `validate_success/1` function?

RTLS (commented):
I'm fine with the refactor. My primary concern is that we will have some regressions in the documentation or dashboard template.

If you want to proceed with this PR I ask that you build an image of the commit and boot the `docker/docker-compose.yaml` then check that the Grafana dashboard continues to work as expected.

This manual QA step would go to great lengths to ensure no regression!

To build the image locally you can do:

```
docker buildx build \
            --platform linux/arm64 \
            --build-arg SELF_HOSTED=1 \
            --build-arg RELEASE_VERSION="dash" \
            --build-arg SENTRY_DSN="https://cd71f754bd50c51be71a5ca7861eacc2@o398678.ingest.us.sentry.io/4508576886489088" \
            --cache-from "type=registry,ref=sequin/sequin:buildcache-arm64" \
            --cache-to "type=registry,ref=sequin/sequin:buildcache-arm64,mode=max" \
            --provenance=false \
            -t sequin:dash \
            . --load
```

then adjust the `docker/docker-compose.yaml` like so:

```
diff --git a/docker/docker-compose.yaml b/docker/docker-compose.yaml
index bf252490..87dc8367 100644
--- a/docker/docker-compose.yaml
+++ b/docker/docker-compose.yaml
@@ -2,7 +2,7 @@ name: sequin

 services:
   sequin:
-    image: sequin/sequin:latest
+    image: sequin:dash
     pull_policy: always
     ports:
       - "7376:7376"
```

PR #1259, opened 2025-04-16
⚡️ [SlotProcessorServer] Disconnect, don't stop, if at system mem limit

acco:
First, there was an if statement that was returning a tuple that was actually unused.

Second, when we've hit system memory limits, we should undergo the same behavior as when payload limit is exceeded: we should disconnect instead of crashing. Disconnecting is less dramatic, as we preserve our `last_flushed_lsn`, which means we can skip messages rapidly on reconnection.

### DIFF
--- a/lib/sequin/runtime/slot_processor_server.ex
+++ b/lib/sequin/runtime/slot_processor_server.ex
@@ -287,20 +287,20 @@ defmodule Sequin.Runtime.SlotProcessorServer do
     current_memory = state.check_memory_fn.()
 
     if current_memory > state.max_memory_bytes do
-      Logger.warning("[SlotProcessorServer] System at memory limit, shutting down",
+      Logger.warning("[SlotProcessorServer] System at memory limit, disconnecting",
         limit: state.max_memory_bytes,
         current_memory: current_memory
       )
 
-      {:disconnect, :over_system_memory_limit}
+      {:disconnect, :over_system_memory_limit, state}
+    else
+      {:stream, query, [],
+       %{
+         state
+         | connection_state: :streaming,
+           low_watermark_wal_cursor: low_watermark_wal_cursor
+       }}
     end
-
-    {:stream, query, [],
-     %{
-       state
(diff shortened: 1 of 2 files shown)


acco:
* **#1259** <a href="https://app.graphite.dev/github/pr/sequinstream/sequin/1259?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/sequinstream/sequin/1259?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

lib/sequin/runtime/slot_processor_server.ex
```diff
@@ -293,14 +293,14 @@ defmodule Sequin.Runtime.SlotProcessorServer do
       )
 
       {:disconnect, :over_system_memory_limit}
+    else
```
yordis, line 296: Call me crazy, this is yet another reason why I use `case` over `if` statements, almost 99% of the time

```elixir
case current_memory > state.max_memory_bytes do
  true -> ...
  false -> ...
end
```

🤷🏻

RTLS (approved):

lib/sequin/runtime/slot_processor_server.ex
RTLS: so currently (before this PR) we are failing to shut down immediately `on_connect` if over system memory limit? this is perhaps related to our issue of exceeding the soft memory limit

RTLS (approved):

lib/sequin/runtime/slot_processor_server.ex
acco: @RTLS That's right! We disconnect when we hit memory limit, then the idea is that when we reconnect after timeout, we should shut off right away if at limit - and we weren't doing that. Meaning we'd accumulate messages until next checkpoint threshold _then_ shut off

PR #1260, opened 2025-04-16
feat: hibernate slot_message_store when all messages are processed

abc3:
### Problem

Under high load, the `SlotMessageStore` process tends to retain large binaries due to message accumulation.  
Because the process stays active, GC is delayed, causing the binary heap to grow uncontrollably.

### Solution

This PR introduces an optimization to hibernate the `SlotMessageStore` process when there are no more messages to handle.  
Hibernation triggers garbage collection and suspends the process until the next message arrives, or until `ConsumerProducer` checks for messages due to `batch_timeout`.

Additionally:
- Introduced a `messages_processed` message to let consumers reset the `receive_timer` once all messages have been delivered.
- Increased the `batch_timeout` from 10 seconds to 1 minute, as it's primarily a fallback. `SlotMessageStore` triggers processing immediately upon receiving new messages.

### DIFF
--- a/lib/sequin/runtime/consumer_producer.ex
+++ b/lib/sequin/runtime/consumer_producer.ex
@@ -46,7 +46,7 @@ defmodule Sequin.Runtime.ConsumerProducer do
       consumer: consumer,
       receive_timer: nil,
       trim_timer: nil,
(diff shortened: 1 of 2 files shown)


acco (approved):
Thanks @abc3 - just tested and helps a lot!

PR #1272, opened 2025-04-19
Typesense reference

fazzone:


docs/reference/sinks/typesense.mdx
```diff
@@ -0,0 +1,76 @@
...
+
+The Typesense sink indexes documents into a Typesense collection, using the Typesense JSON API.
+
+## Configuration
+
+- **Endpoint URL**
+
+    The URL of your Typesense server (e.g., `https://your-typesense-server:8108`).
+
+- **Collection Name**
```
acco, line 15: I switched labels to Sentence case (vs Title Case)

docs/reference/sinks/typesense.mdx
```diff
@@ -0,0 +1,76 @@
...
+
+    The name of the Typesense collection where documents will be indexed.
+    The collection must exist before imports can work, but
```
acco, line 18: - To match style elsewhere, either keep these on the same line or break with an extra blank line between paragraphs (goes for the rest of doc)
- `but ...` Cliffhanger!

docs/reference/sinks/typesense.mdx
```diff
@@ -0,0 +1,76 @@
...
+    The name of the Typesense collection where documents will be indexed.
+    The collection must exist before imports can work, but 
+
+- **API Key**
+
+    The API key for authenticating with your Typesense server.
+
+- **Batch Size** (optional)
+
+    Number of documents to batch together in a single import request to Typesense. Default is 40. Maximum is 10,000.
```
acco, line 26: I'd add a little color on how one chooses the batch size. Either in-line here or as a separate section in this doc (with an anchor link here)

docs/reference/sinks/typesense.mdx
```diff
@@ -0,0 +1,76 @@
...
+
+- **Timeout (seconds)** (optional)
+
+    Number of seconds to wait for a response from Typesense. Default is 5 seconds. Maximum is 300 seconds.
```
acco, line 30: I'd add a little color about aligning timeout with the timeout on the Typesense server (I believe I saw something like this in how-to or quickstart)

docs/reference/sinks/typesense.mdx
```diff
@@ -0,0 +1,76 @@
...
+
+## Transform Requirements
```
acco, line 32:
```suggestion
## Transform requirements
```

docs/reference/sinks/typesense.mdx
```diff
@@ -0,0 +1,76 @@
...
+
+Your [transform](/docs/reference/transforms) must return a document matching the schema of the [Typesense collection](https://typesense.org/docs/28.0/api/collections.html#create-a-collection).
+
+This includes a string `id` field which is mandatory for Typesense.
+It's also okay to use an integer/serial/bigserial/uuid `id` - Sequin will automatically stringify it for you.
+However, from Typesense's perspective, all `id`s are strings and will be returned as such in search results.
+
+Note: the `id` should not include spaces or any other characters that require encoding in urls.
```
acco, line 40: Consider including an example code block

docs/reference/sinks/typesense.mdx
```diff
@@ -0,0 +1,76 @@
...
+
+## Import action
+
+Sequin always uses Typesense's `emplace` import action, which means:
```
acco, line 44: Link to Typesense docs about emplace?

acco: Is this true for deletes as well?

docs/reference/sinks/typesense.mdx
```diff
@@ -0,0 +1,76 @@
...
+
+- Typesense will create a new document or update an existing one based on the `id`
+- Your transform can supply either a complete document or a partial document for update.
+
+Note: The full document should be provided for creation -  partials are only allowed when updating an existing document.
+
+## API endpoints used by Sequin
```
acco, line 51: Great idea to include this section

docs/reference/sinks/typesense.mdx
```diff
@@ -0,0 +1,76 @@
...
+
+Sequin uses the following specific endpoints of the [Typesense documents API](https://typesense.org/docs/28.0/api/documents.html):
+
+- "/collections/#{collection_name}/documents/import"
```
acco, line 55: Use `code` here?

```suggestion
- `/collections/#{collection_name}/documents/import`
```

docs/reference/sinks/typesense.mdx
```diff
@@ -0,0 +1,76 @@
...
+  For indexing batches of documents.  This is the most common path.
+- "/collections/#{collection_name}/documents"
+  For indexing documents one at a time.
+  Used when batch size was manually configured to be 1, or when the batch happens to have only 1 document.
+- "/collections/#{collection_name}/documents/#{document_id}"
+  For deleting documents.  Deletions are always processed one at a time.
+
+Sequin does not interact with the collections API, or perform any searches.
+
+## Error handling
```
acco, line 65: Great!

docs/reference/sinks/typesense.mdx
fazzone: I feel as though I don't myself know how to choose a good batch size.  Is there any deliberate buffering within sequin to enable larger batches, like Nagle's algorithm does for tcp?  If not, then I think a batch size of 1 would be the most common case in practice, since that's probably what would happen if you change one row and there isn't a backlog of replication messages.  In any case, the batch size is probably much more important to the backfill than to steady-state operation

docs/reference/sinks/typesense.mdx
fazzone: 👀

docs/reference/sinks/typesense.mdx
fazzone: For deletes we get the id from the transformed message but don't care about or send any other part

docs/reference/sinks/typesense.mdx
fazzone: 👍

docs/reference/sinks/typesense.mdx
fazzone: 👍

docs/reference/sinks/typesense.mdx
fazzone: For code style as opposed to how it looks right? I was thinking by markdown rules that single line breaks in the source text don't change the rendered output

docs/reference/sinks/typesense.mdx
acco: We use Broadway's batching. Batch size probably only makes sense in the context of "Batch timeout" or whatever – we want to make these available on sinks (some of them have hard-coded settings):

https://github.com/sequinstream/sequin/blob/master/lib/sequin/runtime/sink_pipeline.ex#L250-L266

So, if there's nothing more to add, we can skip for now. When we go back and add "Batch timeout" everywhere, we can add some color.

Or, it might be one of those advanced settings that we try to control for you with in most places. For example, it's reasonable to set a pretty low batch timeout (say 100ms) and max batch size, and just try to fill up as big of a batch as we can in a short window before shipping out

fazzone: So batch_timeout is exactly such a thing which waits to fill a batch before sending an incomplete one?

docs/reference/sinks/typesense.mdx
acco: For code style! Was not sure about rendered output

docs/reference/sinks/typesense.mdx
acco: Yes, send batch when either (a) batch is full or (b) timeout has expired

acco (approved):

PR #1277, opened 2025-04-19
🐛 Improve parsing of pg arrays

acco:
PR 1 / 2

2nd PR extracts the value casting out so we can unit test everything

### DIFF
--- a/lib/sequin/postgres/array_lexer.ex
+++ b/lib/sequin/postgres/array_lexer.ex
@@ -1,0 +1,129 @@
+defmodule Sequin.Postgres.ArrayLexer do
+  @moduledoc """
+  Lexer for Postgres array literals.
+
+  * **Splits only** – every scalar comes back as a *string* (or `nil` for the
+    literal `NULL`). No JSON decoding or type‑casting takes place here.
+  * **Nesting preserved** – any inner arrays are returned as nested Elixir
+    lists.
+  * **Return contract** – always `{:ok, result}` on success or
+    `{:error, reason}` on failure.
+
+  ### Examples
+
+      iex> Sequin.Postgres.ArrayLexer.lex("{}")
+      {:ok, []}
+
+      iex> Sequin.Postgres.ArrayLexer.lex("{1,2,3}")
+      {:ok, ["1", "2", "3"]}
+
+      iex> Sequin.Postgres.ArrayLexer.lex(~s({"a","b","c"}))
+      {:ok, ["a", "b", "c"]}
+
+      iex> Sequin.Postgres.ArrayLexer.lex("{{1,2},{3,4}}")
+      {:ok, [["1", "2"], ["3", "4"]]}
+  """
+  alias Sequin.Error
+
+  @type element :: String.t() | nil | [element]
+  @type result :: {:ok, [element]} | {:error, term}
+
+  @spec lex(String.t()) :: result
+  def lex(binary) when is_binary(binary) do
+    binary = String.trim(binary)
+
+    case parse_array(binary) do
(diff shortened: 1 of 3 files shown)


acco:
* **#1277** <a href="https://app.graphite.dev/github/pr/sequinstream/sequin/1277?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/sequinstream/sequin/1277?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

lib/sequin/postgres/array_lexer.ex
```diff
@@ -0,0 +1,127 @@
...
+    {:ok, Enum.reverse(put_buf(buf, acc)), rest}
+  end
+
+  # Comma between elements
+  defp parse_items(<<",", rest::binary>>, acc, buf) do
+    parse_items(rest, put_buf(buf, acc), nil)
+  end
+
+  # Ignore whitespace outside of quotes
+  defp parse_items(<<c, rest::binary>>, acc, buf) when c in [32, 9, 10, 13] do
```
fazzone, line 66: I think you can write e.g. `@ws String.to_charlist(" \r\n\t")` and then use `@ws` in a guard

lib/sequin/postgres/array_lexer.ex
```diff
@@ -0,0 +1,127 @@
...
+  # Quoted element
+  defp parse_items(<<"\"", rest::binary>>, acc, nil) do
+    with {:ok, value, rest2} <- parse_quoted(rest) do
+      parse_items(rest2, acc, value)
+    end
+  end
+
+  # Backslash escape (outside quotes) → keep the next char literally
+  defp parse_items(<<"\\", c, rest::binary>>, acc, buf) do
+    parse_items(rest, acc, (buf || "") <> <<c>>)
```
fazzone, line 86: does buf=nil mean something special?  why not use buf="" from the start?

fazzone (approved):

lib/sequin/postgres/array_lexer.ex
acco: Good idea!

lib/sequin/postgres/array_lexer.ex
```diff
@@ -0,0 +1,127 @@
...
+
+  ## helpers -----------------------------------------------------------------
+
+  # Adds the finished buffer (if any) to the accumulator.
+  # no element captured
+  defp put_buf(nil, acc), do: acc
+
+  defp put_buf(buf, acc) when is_binary(buf) do
+    elem =
+      case String.trim(buf) do
```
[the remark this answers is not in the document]
acco: I don't think there will be, I'll remove

lib/sequin/postgres/array_lexer.ex
acco: buf=nil and buf="" do have two distinct meanings:

buf=nil --> nothing in buffer, at start of array { or just encountered a ,
buf="" --> could be the middle element here:

`{"a", "", "b"}`

Hence, we want buf to start out as nil so we can tell the difference.
  
  The alt way to write this would be:
  
  ```elixir
  # First char into the buffer
  defp parse_items(<<c, rest::binary>>, acc, nil) do
    parse_items(rest, acc, <<c>>)
  end
  
  # Append char to buffer
  defp parse_items(<<c, rest::binary>>, acc, buf) do
    parse_items(rest, acc, buf <> <<c>>)
  end
  ```
  
Perhaps a little clearer, so I'll propagate!

PR #1280, opened 2025-04-20
⚡️ Add retry attempts when system is at memory limit

acco:
When the system hits the memory limit, right now we disconnect right away. By instead giving the system up to 3 seconds to drain below the limit, we can dramatically improve performance in situations where we are near the limit.

The ideal fix is to not pull messages into the system anyway when we're at the limit, but that improvement can come later.

### DIFF
--- a/lib/sequin/runtime/slot_processor_server.ex
+++ b/lib/sequin/runtime/slot_processor_server.ex
@@ -1732,13 +1732,23 @@ defmodule Sequin.Runtime.SlotProcessorServer do
     conn
   end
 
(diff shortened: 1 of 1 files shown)


acco:
* **#1280** <a href="https://app.graphite.dev/github/pr/sequinstream/sequin/1280?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/sequinstream/sequin/1280?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1286, opened 2025-04-21
Fix comment/code in table_reader_server task

fazzone:
It seems like a fine idea but I don't think was doing what it said


acco (approved):

acco:
Nice catch!

PR #1289, opened 2025-04-21
minielixir: don't explode when we see an error we've never seen before

fazzone:
On the subject of protocols, note that you can effectively call Sequin's Jason.Encoder impl for a struct by name because constructing structs is allowed.  However, a Jason.Encoder impl which is somehow dangerous when called with default values is a bad idea for other reasons.


fazzone:
superseded by #1290

PR #1292, opened 2025-04-22
 typesense: remove batch size & timeout from UI, docs 

fazzone:


acco (approved):
Looks good -- looks like Typesense sets its defaults for batch config in `Sequin.Runtime.TypesensePipeline` yeah?

PR #1324, opened 2025-04-25
cli: print diagnostic message from config export to stderr

fazzone:
This means that you can run e.g. `./sequin config export > sequin.yaml`


RTLS (approved):
Great improvement- using this in a demo tomorrow 💪 @fazzone

Issue #1327, opened 2025-04-25
After `messages_success`, check if we have unblocked any messages and publish


acco:
* Check if any other messages exist for the group_ids that we just popped (just succeeded)

PR #1334, opened 2025-04-28
Check for recently unblocked messages after ack

acco:
SMS broadcasts when we ingest new messages. This prompts ConsumerProducer to schedule a pull to check for new messages.

However, there's another scenario where messages will be made available: when messages are *acked*. In that situation, a message that was previously blocked may be available for pulling.

So, we broadcast when that happens as well. Otherwise, you can see a weird pattern where a message belonging to the same group ID as a previous messages takes several seconds to deliver after its counterpart delivers.

Addresses #1327

### DIFF
--- a/lib/sequin/runtime/consumer_producer.ex
+++ b/lib/sequin/runtime/consumer_producer.ex
@@ -39,7 +39,7 @@ defmodule Sequin.Runtime.ConsumerProducer do
       Mox.allow(Sequin.Runtime.SlotMessageStoreMock, test_pid, self())
     end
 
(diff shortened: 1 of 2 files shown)


acco:
* **#1334** <a href="https://app.graphite.dev/github/pr/sequinstream/sequin/1334?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/sequinstream/sequin/1334?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1364, opened 2025-05-05
feat: add ExpandableCard component and use it in SinkConsumerForm

### LINKED ISSUE #1288: For drawers, make entire title bar clickable
This whole area should be clickable to expand/collapse, not just the arrow:

davoclavo:
Hello!

Here we add an `ExpandableCard` component which handles the proper UI behavior for these type of cards. And implement usages of it in SinkConsumerForm, and GroupColumnsForm. This fixes issue #1288.

Additional notes:

- To avoid confusion, now the ChevronDown is hidden from the Card title if the card is in disabled state, such as when no table is selected when creating or editing a sink
- Also fixes a bug where if a table is not selected, it didn't show the "Please select a table first." message in each section.
- Because the `ExpandableCard` component may show a summary or a content fragment depending on the state of the card, I opted to use named slots via svelte:fragment tags, instead of a single slot that handles those state variants. Hopefully thats fine.
- I had to add some hacky `<a on:click|stopPropagation>` wrappers to the Popovers to avoid expanding the cards when clicking on the the Popover info icon located in some card titles. Probably would be best to add this to a custom Popover component, or suggest that improvement in bits-ui, but perhaps for another time.
  - There is a small quirk (existing prior to this hack) where if you click the Popover it takes you to the top of the page. Didn't have time to look into that but definitely something that can be polished later.

Any feedback is more than welcome!

---

# Dynamic Evidence

## Gif

image

# Static Evidence

### Scenario 1: No table selected
image

### Scenario 2: Table selected, all components collapsed by default
image

### Scenario 3: Table selected, all components expanded after clicking header
image

### Scenario 1: Transform card collapsed by default
image

### Scenario 2: Transform card expanded after clicking header
image

### DIFF
--- a/assets/svelte/components/ui/card/expandable-card.svelte
+++ b/assets/svelte/components/ui/card/expandable-card.svelte
@@ -1,0 +1,44 @@
+<script lang="ts">
+  import { Card, CardContent, CardHeader, CardTitle } from ".";
+  import { ChevronDown } from "lucide-svelte";
+
+  export let expanded: boolean = false;
+  export let disabled: boolean = false;
+
+  function handleClick() {
+    if (disabled) return;
+    expanded = !expanded;
+  }
+</script>
+
+<Card>
+  <CardHeader class="flex flex-row items-center justify-between">
+    <button
+      type="button"
+      class="flex flex-row items-center justify-between w-full"
+      on:click={handleClick}
+      {disabled}
+    >
+      <CardTitle class="flex items-center gap-2">
+        <slot name="title" />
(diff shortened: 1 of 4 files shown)


davoclavo:
By the way... in Sink Creation I would suggest to have all cards be expanded by default, and keeping them collapsed by default in Sink Edition.

I say this because I feel it is a bit unclear you are able to expand them when you aren't familiar with Transforms, Backfill and Groupings - given that a short summary of the default option is selected.

PR #1373, opened 2025-05-06
chore: update type specs, remove dead code, and improve url validation

abc3:


lib/sequin/consumers/elasticsearch_sink.ex
```diff
@@ -29,11 +29,8 @@ defmodule Sequin.Consumers.ElasticsearchSink do
     changeset
     |> validate_change(:endpoint_url, fn :endpoint_url, url ->
       case URI.parse(url) do
-        %URI{} ->
-          []
-
-        _ ->
-          [endpoint_url: "must be a valid URL"]
+        %URI{scheme: scheme, host: host} when not is_nil(scheme) and not is_nil(host) -> []
+        _ -> [endpoint_url: "must be a valid URL"]
```
abc3, line 33: `URI.parse/1` always returns `%URI{}`, so assuming that a url is valid if at least the scheme and host are not `nil`

lib/sequin/health/check_http_endpoint_health_worker.ex
```diff
@@ -17,14 +17,8 @@ defmodule Sequin.Health.CheckHttpEndpointHealthWorker do
   @impl Oban.Worker
   def perform(%Oban.Job{args: %{"http_endpoint_id" => http_endpoint_id}}) do
     case Consumers.get_http_endpoint(http_endpoint_id) do
```
abc3, line 19: get_http_endpoint returns only {:ok, HttpEndpoint.t()} or {:error, NotFoundError.t()}

lib/sequin/redis.ex
```diff
@@ -83,7 +83,6 @@ defmodule Sequin.Redis do
 
       case res do
         {:ok, result} -> result
-        {:error, error} when is_exception(error) -> raise error
         {:error, error} -> raise Error.service(service: :redis, code: :command_failed, message: error)
```
abc3, line 86: handling the same behavior

lib/sequin/runtime/kafka_pipeline.ex
```diff
@@ -107,11 +107,6 @@ defmodule Sequin.Runtime.KafkaPipeline do
       {:ok, partition_count} ->
         partition_count
 
-      {:error, error} when is_atom(error) ->
-        error = Error.invariant(message: "Failed to get partition count for Kafka sink: #{inspect(error)}")
-        Logger.warning("[KafkaPipeline] #{Exception.message(error)}")
-        raise error
```
abc3, line 113: there is no error with atom type

PR #1374, opened 2025-05-06
feat: show clearer connectivity errors and run extra checks for IPv6

### LINKED ISSUE #1097: If the db uses ipv6 but the host machine doesn't support, show error
Some dbs, like Supabase, use ipv6. We detect if the db uses ipv6 or not. But we don't detect if the host machine has ipv6 support.

We can improve the error here by surfacing this.

Anywhere that we call `NetworkUtils.check_ipv6` is a candidate - in fact, arguably, that function should return an error if the target requires ipv6 but the host machine doesn't support.

davoclavo:
Show clearer error messages when performing checks for IPv6 hosts, and also perform additional check if machine has any IPv6 assigned to any network interface as a way to determine machine support.

Notes:
 - I have only tested with ISP/networks not supporting IPv6, and manually disabling IPv6 for my network interface, however and I am actively looking for a way to test this with a machine with actual full non-IPv6 support. Any ideas are welcome.
 - Another additional approach would be to try to open a socket and bind it to a IPv6 address, but I am yet unsure if it provides any benefit over checking network interface support.
 - All tests were performed in the Connect Database page
 - In the case of performing a connectivity test to an IPv6 host, we need to mainly catch the `:ehostunreach` - so I grouped it with the existing `:nxdomain` error case (which should not ever happen I think, as we already did DNS resolutions and determined it is an IPv6 host)
 - I opted to implement the additional check in `NetworkUtils.test_tcp_reachability` rather than the suggested `NetworkUtils.check_ipv6` mainly because even if your machine supports it, your ISP/network may not, and we can only check that by actually opening a TCP socket to the given ip+port - also because UI errors are mainly surfaced from the `test_tcp_reachability`, and would require more piping to forward them from `check_ipv6`

Fixes #1097

----

# Evidence

**Before**
image

**After**
When your machine seems to support IPv6 but not your ISP/network
image

When your machine seems to not support IPv6
image

### DIFF
--- a/lib/sequin/network_utils.ex
+++ b/lib/sequin/network_utils.ex
@@ -20,7 +20,7 @@ defmodule Sequin.NetworkUtils do
   end
 
   @doc """
-  Pings a host on a port to see if it is reachable.
+  Establishes a test TCP connection to a host on a port to see if it is reachable.
   """
   @spec test_tcp_reachability(host :: String.t(), port :: number(), ipv6 :: boolean(), timeout :: number()) ::
           :ok | {:error, Error.t()}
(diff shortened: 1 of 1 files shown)


acco:
Great work here @davoclavo , and thanks for the detailed report! This will be a great improvement 🚀

PR #1387, opened 2025-05-07
fix: avoid scrolling to top whenever a popover is closed within a full page modal

### LINKED ISSUE #724: Get forms out of FullPageModal
Right now, `consumer_form` is a LiveComponent. We render it inside of the `FullPageModal` component because:

* A full page modal is a nice UX pattern for displaying a form for create or edit: mentally, the user is still on the page where they clicked "Create/Edit"
* We don't want to display the sidenav when displaying a form.

This works mostly well - however, there are some quirks. Whenever a tooltip is closed, the window scrolls to the top of the FullPageModal - not desirable.

We either want to:

1. Determine why the scroll top behavior occurs - and fix it (we're using ShadCN Svelte).
2. Move `consumer_form` into a LiveView. Just redirect to a different path when rendering the form. It should still *feel* like a full-page modal, but it won't be.

davoclavo:
Fix the quirk where the page scrolls to top whenever a popover is closed within a full page modal

This is a **very** ugly hack, but after a couple of hours trying to find the underlying root cause, I was unable to, and this seems to reliably fix the issue while not including any other detectable issue (other than showing the exit dialog for about 20ms).

Hopefully we can get to the bottom of this at some point, but this could be a valid workaround for the time being.

Fixes #724

### DIFF
--- a/assets/svelte/components/FullPageModal.svelte
+++ b/assets/svelte/components/FullPageModal.svelte
@@ -39,6 +39,24 @@
   onMount(() => {
     window.addEventListener("keydown", handleEscapeKey);
     listening = true;
+
+    // NOTE: This is an ugly hack to avoid page being scrolled to the top whenever a Popover is closed within a Dialog window
+    window.setTimeout(() => {
+      window.dispatchEvent(new KeyboardEvent("keydown", { key: "Escape" }));
+      window.setTimeout(() => {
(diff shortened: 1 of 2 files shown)


davoclavo:
superseded by https://github.com/sequinstream/sequin/pull/1395

PR #1395, opened 2025-05-08
fix: improve full page forms to avoid using a modal

### LINKED ISSUE #724: Get forms out of FullPageModal

davoclavo:
Improve rendering full page forms by removing the Modal/Dialog wrapper that was causing some quirks with some event handling.

Forms:
 - Sink
 - Database
 - HTTP endpoint
 - WAL Pipeline

Related additional improvements:
 - The new component name is changed from `FullPageModal` to `FullPageForm`
   - Simplified external props
 - Now we ask for confirmation on exit by default, instead of the opposite. It is still configurable via a prop if needed to override.
 - Replaced read only `Popovers` for `Tooltips`. Keep Popovers for cases where we need an additional input. This improves usability as information is shown on hover.
 - Fixed Message Group card on Sink form to be expanded by default
 - Improved rendering the horizontal scrollbar on `CodeWithCopy` blocks

Shoutout to @acco for the help figuring a robust solution!

Fixes #724

### DIFF
--- a/lib/sequin_web/live/wal_pipelines/form.ex
+++ b/lib/sequin_web/live/wal_pipelines/form.ex
@@ -24,13 +24,15 @@ defmodule SequinWeb.WalPipelinesLive.Form do
 
     case fetch_or_build_wal_pipeline(socket, wal_pipeline_id) do
       {:ok, %WalPipeline{} = wal_pipeline} ->
-        {:ok,
-         socket
-         |> assign(wal_pipeline: wal_pipeline)
-         |> assign(is_edit: is_edit)
-         |> assign(show_errors?: false)
-         |> assign(submit_error: nil)
-         |> assign_changeset(%{}), layout: {SequinWeb.Layouts, :app_no_sidenav}}
+        socket =
+          socket
+          |> assign(wal_pipeline: wal_pipeline)
+          |> assign(is_edit: is_edit)
+          |> assign(show_errors?: false)
+          |> assign(submit_error: nil)
+          |> assign_changeset(%{})
+
+        {:ok, socket, layout: {SequinWeb.Layouts, :app_no_sidenav}}
 
       {:ok, nil} ->
         {:ok, push_navigate(socket, to: ~p"/change-capture-pipelines"), layout: {SequinWeb.Layouts, :app_no_sidenav}}
--- a/lib/sequin_web/live/wal_pipelines/form.ex
+++ b/lib/sequin_web/live/wal_pipelines/form.ex
@@ -30,10 +30,10 @@ defmodule SequinWeb.WalPipelinesLive.Form do
          |> assign(is_edit: is_edit)
          |> assign(show_errors?: false)
          |> assign(submit_error: nil)
-         |> assign_changeset(%{})}
+         |> assign_changeset(%{}), layout: {SequinWeb.Layouts, :app_no_sidenav}}
 
       {:ok, nil} ->
-        {:ok, push_navigate(socket, to: ~p"/change-capture-pipelines")}
+        {:ok, push_navigate(socket, to: ~p"/change-capture-pipelines"), layout: {SequinWeb.Layouts, :app_no_sidenav}}
     end
   end
 
--- a/assets/svelte/components/CodeWithCopy.svelte
+++ b/assets/svelte/components/CodeWithCopy.svelte
@@ -11,10 +11,10 @@
 </script>
 
 <div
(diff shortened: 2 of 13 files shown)


lib/sequin_web/live/wal_pipelines/form.ex
```diff
@@ -30,10 +30,10 @@ defmodule SequinWeb.WalPipelinesLive.Form do
          |> assign(is_edit: is_edit)
          |> assign(show_errors?: false)
          |> assign(submit_error: nil)
-         |> assign_changeset(%{})}
+         |> assign_changeset(%{}), layout: {SequinWeb.Layouts, :app_no_sidenav}}
```
fazzone, line 33: can you write socket=... on another line above?

fazzone:
The code changes are looking really good!  Loving the net-red diff!  However, trying to test this locally I noticed a weird issue with the test connection button of the create sink form - it looks like the popover is flashing into and out of existence very rapidly when it should be opened until dismissed?  Hard to explain - see the video: https://www.loom.com/share/1aeb311eb6904ad6ab3a2b781f9ee00f?sid=c3978947-4aea-4798-bc0e-e4aec57161cf

davoclavo:
@fazzone thanks a lot for reporting that issue. Let me take a look at it and issue a fix

lib/sequin_web/live/wal_pipelines/form.ex
davoclavo: @fazzone Sure! just to confirm, do you mean to make your suggested change in the render function instead?

```diff
--- a/lib/sequin_web/live/wal_pipelines/form.ex
+++ b/lib/sequin_web/live/wal_pipelines/form.ex
@@ -52,6 +52,8 @@ defmodule SequinWeb.WalPipelinesLive.Form do
     <div id="wal-pipeline-form">
       <.svelte
         name="wal_pipelines/Form"
+        socket={@socket}
         props={
           %{
             walPipeline: encode_wal_pipeline(assigns),
```

Also, would it make sense to add `ssr={false}` as well? I am still unsure of which components should be server side rendered and which ones don't, but I noticed that the databases/form.ex is also setting that property as well

davoclavo:
@fazzone I just pushed a fix to the issue you reported. Thanks for reporting that bug! it was due to my change on the previous `<Popover.Trigger>` structure.

lib/sequin_web/live/wal_pipelines/form.ex
davoclavo: Ohh I think I understand what you mean: to refactor the assignment to not have so many nested brackets/tuples.

```diff
--- a/lib/sequin_web/live/wal_pipelines/form.ex
+++ b/lib/sequin_web/live/wal_pipelines/form.ex
@@ -24,13 +24,15 @@ defmodule SequinWeb.WalPipelinesLive.Form do

     case fetch_or_build_wal_pipeline(socket, wal_pipeline_id) do
       {:ok, %WalPipeline{} = wal_pipeline} ->
-        {:ok,
-         socket
-         |> assign(wal_pipeline: wal_pipeline)
-         |> assign(is_edit: is_edit)
-         |> assign(show_errors?: false)
-         |> assign(submit_error: nil)
-         |> assign_changeset(%{}), layout: {SequinWeb.Layouts, :app_no_sidenav}}
+        socket =
+          socket
+          |> assign(wal_pipeline: wal_pipeline)
+          |> assign(is_edit: is_edit)
+          |> assign(show_errors?: false)
+          |> assign(submit_error: nil)
+          |> assign_changeset(%{}), layout: {SequinWeb.Layouts, :app_no_sidenav}
+
+        {:ok, socket}

```

Aside from that, do you happen to have context on the `socket={@socket}` and `ssr={false}` props I mentioned?

davoclavo: Done!

acco:
This looks great! Awesome work here @davoclavo this has been bugging us for a while 🙏

Issue #1367, opened 2025-05-05
Accept `_{var}` in mini-elixir function headers


acco:
This should be valid mini-elixir:

```
def transform(action, record, changes, _metadata) do
```

(note `_metadata`)

acco:
Fixed in #1410

PR #1410, opened 2025-05-09
feat: accept _{var} in mini-elixir function headers

fahchen:
Addressed: https://github.com/sequinstream/sequin/issues/1367

### DIFF
--- a/lib/sequin/transforms/minielixir/validator.ex
+++ b/lib/sequin/transforms/minielixir/validator.ex
@@ -4,6 +4,7 @@ defmodule Sequin.Transforms.MiniElixir.Validator do
   @args [:action, :record, :changes, :metadata]
   @error_bad_toplevel "Expecting only `def transform` or `def route` at the top level"
   @error_invalid_name "Only function names `transform` or `route` are allowed"
+  @error_bad_args "The parameter list `#{Enum.join(@args, ", ")}` is required"
 
   def create_expr(body_ast, modname) do
     :ok = check(body_ast)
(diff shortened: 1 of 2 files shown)


fazzone (approved):
love it!  very elixir 💪

lib/sequin/transforms/minielixir/validator.ex
```diff
@@ -39,11 +40,27 @@ defmodule Sequin.Transforms.MiniElixir.Validator do
...
+    end
+  end
+
+  for arg <- @args do
+    defp unwrap_params([{unquote(arg), _meta, _context} | rest], acc) do
+      unwrap_params(rest, [unquote(arg) | acc])
     end
+
+    defp unwrap_params([{unquote(:"_#{arg}"), _meta, _context} | rest], acc) do
+      unwrap_params(rest, [unquote(arg) | acc])
```
fazzone, line 58: I think the result of this will be that whether you wrote `metadata` or `_metadata`, it will always behave as if you had written `metadata`.  Right now the warning for an unused variable isn't propagated to the user, so they appear the same from the outside.

What do you think about having `[arg, :"_#{arg}", :"_"]` as a second generator clause in the for, instead of defining the two clauses in the for body?

fahchen: > I think the result of this will be that whether you wrote metadata or _metadata, it will always behave as if you had written metadata. Right now the warning for an unused variable isn't propagated to the user, so they appear the same from the outside.

i also noticed that we just pass the four variables as binding when evaling the code. which means, `metadata` always be assigned a value even if we write it to `_metadata` or whatever else. the named variables in the function header seems not to work as expected.

> What do you think about having [arg, :"_#{arg}", :"_"] as a second generator clause in the for, instead of defining the two clauses in the for body?

sounds great, but I'm a bit worried about overthinking, because it's my first contribution here. i'm just following whatever the issue defines. btw, i can draft another pr, if you'd like.

PR #1420, opened 2025-05-10
fix: Message Grouping card form improvements

### LINKED ISSUE #1383: Clearly show "Message grouping" as disabled in edit state
### Context

When a user edits an existing **sink**, the **Message Grouping** card shows in a read-only state. We should make it even clearer that Message grouping is not editable, and that you must create a new sink to modify this behavior.

### Problem

* There is no visual cue or explanatory text telling the user why the field can’t be changed.

### Proposed solution

1. **Chevron remains visible** (to keep the layout consistent).
2. Hovering on the Message Grouping card title bar:
   * A tooltip appears:

     > “Message grouping can’t be changed for an active sink.”
3. The dropdown cannot be opened.

davoclavo:
Improve Message Grouping card with the following changes:

- Always show this card as an ExpandableCard, in order to show the Chevron and keep the layout consistent
- For the Sink Edit form show tooltip with the notice "Message grouping can’t be changed for an active sink." on hover on the entire title bar
- Change default expanded state whenever the card is shown in Sink Creation

Fixes #1383

---

# Evidence

## Sink Create form: expanded by default and no tooltip
image

## Sink Edit form: show Tooltip on hover + show disabled Chevron
image

### DIFF
--- a/assets/svelte/components/ui/card/expandable-card-header.svelte
+++ b/assets/svelte/components/ui/card/expandable-card-header.svelte
@@ -1,0 +1,34 @@
+<script lang="ts">
+  import { CardHeader, CardTitle } from ".";
+  import { ChevronDown } from "lucide-svelte";
+  import { Button as ButtonPrimitive } from "bits-ui";
+
+  export let expanded: boolean = false;
+  export let disabled: boolean = false;
+  export let builders: $$Props["builders"] = [];
+
+  function handleClick() {
+    if (disabled) return;
+    expanded = !expanded;
+  }
+</script>
+
+<CardHeader class="flex flex-row items-center justify-between">
+  <ButtonPrimitive.Root
+    class="flex flex-row items-center justify-between w-full cursor-pointer"
+    on:click={handleClick}
+    {builders}
+    {disabled}
+  >
+    <CardTitle class="flex items-center gap-2">
+      <slot />
+    </CardTitle>
+    <div
+      class="transition-transform duration-200"
+      class:rotate-180={expanded}
+      class:text-gray-300={disabled}
+    >
+      <ChevronDown class="h-4 w-4" />
+    </div>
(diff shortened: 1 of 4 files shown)


acco:
@davoclavo This looks good to me! One minor change: Regarding this:

> Change default expanded state whenever the card is shown in Sink Creation

We actually purposefully show it as collapsed on create, as users rarely change/deviate from the default on this one. That may change in the future. (We think 80%+ will want us to group by primary key)

Note if the table does _not_ have any primary keys, then I believe it will be expanded –– as there is no default behavior. Might be worth checking what the behavior is there!

davoclavo:
@acco Thanks for the feedback! 

I just pushed the changes for the behavior you pointed out, and also identified a couple of extra improvements.

- Message grouping card
  - In **Sink Create** mode
    - For tables **with** primary keys the card is shown as collapsed by default, as you suggested, and expanding is available.
    - For tables **without** primary keys the card is shown as expanded by default, in order to show the option to select which column to group on. Also disabled the collapsing behavior from the card as a column **must** be selected. 
  - In **Sink Edit** mode
    - Card is always collapsed and disabled. Also shows the tooltip stating that grouping can only be done when a sink is created
- Made a change on `ExpandableCards` to allow an extra customization of disabling a card in either collapsed or expanded state. Previously a disabled card forced the state to be collapsed.
- Added `type="button"` to the `ExpandableCardHeader` component, otherwise was causing some UX issues like submitting the form or showing form validation results unexpectedly.

---

# Evidence

## Sink Create - using table **with** primary keys

image


## Sink Create - using table **without** primary keys


image

## Sink Edit - using table **with** primary keys
image

## Sink Edit - using table **without** primary keys

image

acco:
@davoclavo Awesome work, all sounds good!

Issue #1298, opened 2025-04-22
Improve handling of secrets in `sequin.yaml`


acco:
For e.g. database passwords, AWS API tokens, and encrypted HTTP headers, we can improve the YAML experience.

**Imports**

During imports, there are situations where we don't show you the diff if it's for a sensitive field. That's because the `Transforms.to_external` omits sensitive fields. And we [perform the comparison between the external structures](https://github.com/sequinstream/sequin/blob/master/lib/sequin_web/controllers/yaml_controller.ex#L154-L170).

We'll need to fix this. It might be fine to leak the old sensitive field — we should verify what Terraform does in this instance.

**Exports**

During an export, encrypted fields should not be exported by default — they should be ignored. We should show a warning of such at the bottom of the export output (the export output contains warnings/next steps).

We should add a flag `--include-sensitive` that exports with sensitive fields.

PR #1432, opened 2025-05-12
fix: improve yaml load for `sinks.max_retry_count`, `sinks.active_backfill` and `http_endpoints.encrypted_headers`

davoclavo:
While testing out CLI export->plan functionality, I detected a couple of issues when trying to load a config from a yaml.

1. `sinks.max_retry_count` - if is set to nil (as it seems its the default) the loader fails with an error stating that it should be greater than 0
2. `sinks.active_backfill` - if is present, and set to any value, the loader fails. I suggest to ignore this field for now until we support managing backfills via the yaml.
3. `http_endpoints.encrypted_headers` - seems to be [serialized as a string/binary](https://github.com/sequinstream/sequin/blob/main/lib/sequin/transforms/transforms.ex#L418-L420) (even if the map is empty, as it seems is the default) - causing to fail `parse_headers` validation. I suggest ignoring this field for now, unless we want to handle encrypting/decrypting them properly
  3.1. I also made a change to match if encrypted headers is an empty map, as opposed to `nil` and serialize them as an empty map 

I added a simple test that includes a single resource of each, mostly created with default settings. 

For a future iteration of an improved but similar check I would suggest to write a test that creates a bunch of records and combinations (`databases`, `change_retentions`, `sinks`, `http_endpoint`, `functions`); then assert that exporting->loading works. That way we can test that if the exporting/serializing logic changes, it is still compatible with the loading/deserializing logic and viceversa.

### DIFF
--- a/lib/sequin/transforms/transforms.ex
+++ b/lib/sequin/transforms/transforms.ex
@@ -403,6 +403,10 @@ defmodule Sequin.Transforms do
 
   defp encrypted_headers(%HttpEndpoint{encrypted_headers: nil}), do: %{}
 
+  defp encrypted_headers(%HttpEndpoint{encrypted_headers: encrypted_headers})
+       when is_map(encrypted_headers) and map_size(encrypted_headers) == 0,
+       do: %{}
+
   defp encrypted_headers(%HttpEndpoint{encrypted_headers: encrypted_headers}) do
     "(#{map_size(encrypted_headers)} encrypted header(s)) - sha256sum: #{sha256sum(encrypted_headers)}"
   end
--- a/lib/sequin/transforms/transforms.ex
+++ b/lib/sequin/transforms/transforms.ex
@@ -431,11 +431,9 @@ defmodule Sequin.Transforms do
             {:error, error} -> {:halt, {:error, error}}
           end
 
+        # Ignore encrypted_headers until encryption/decryption works
         "encrypted_headers" ->
-          case parse_headers(value) do
-            {:ok, headers} -> {:cont, {:ok, Map.put(acc, :encrypted_headers, headers)}}
-            {:error, error} -> {:halt, {:error, error}}
-          end
+          {:cont, {:ok, acc}}
 
         "webhook.site" ->
           if value in [true, "true"] do
@@ -597,6 +595,9 @@ defmodule Sequin.Transforms do
         "max_retry_count" when is_integer(value) and value >= 0 ->
           {:cont, {:ok, Map.put(acc, :max_retry_count, value)}}
 
+        "max_retry_count" when is_nil(value) ->
+          {:cont, {:ok, acc}}
+
         "max_retry_count" ->
           {:halt, {:error, Error.validation(summary: "max_retry_count must be a non-negative integer")}}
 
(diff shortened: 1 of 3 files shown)


davoclavo:
I just noticed that some tests are failing that are testing proper loading for `encrypted_headers` - however I don't currently see a case where they could be serialized like that. 

What is the current state of handling `encrypted_headers`??
If we don't support them for now, is it OK if I adjust these tests to reflect that they will be ignored from load for now?

Example test https://github.com/sequinstream/sequin/blob/main/test/sequin/yaml_loader_test.exs#L343

lib/sequin/transforms/transforms.ex
```diff
@@ -403,6 +403,10 @@ defmodule Sequin.Transforms do
 
   defp encrypted_headers(%HttpEndpoint{encrypted_headers: nil}), do: %{}
 
+  defp encrypted_headers(%HttpEndpoint{encrypted_headers: encrypted_headers})
+       when is_map(encrypted_headers) and map_size(encrypted_headers) == 0,
+       do: %{}
+
```
acco, line 409: Nice addition

lib/sequin/transforms/transforms.ex
```diff
@@ -431,11 +435,9 @@ defmodule Sequin.Transforms do
             {:error, error} -> {:halt, {:error, error}}
           end
 
+        # Ignore encrypted_headers until encryption/decryption works
         "encrypted_headers" ->
-          case parse_headers(value) do
-            {:ok, headers} -> {:cont, {:ok, Map.put(acc, :encrypted_headers, headers)}}
-            {:error, error} -> {:halt, {:error, error}}
-          end
+          {:cont, {:ok, acc}}
```
acco, line 440: I'm going to revert this before merging.

In our `sequin config export`, we mention that some fields are not supported - this is one of them. You're right, we export is as a binary, just a sha of the secret. This is not ideal, but users are definitely using this field for their imports.

We want to improve secret handling here: https://github.com/sequinstream/sequin/issues/1298

lib/sequin/transforms/transforms.ex
```diff
@@ -597,6 +599,9 @@ defmodule Sequin.Transforms do
         "max_retry_count" when is_integer(value) and value >= 0 ->
           {:cont, {:ok, Map.put(acc, :max_retry_count, value)}}
 
+        "max_retry_count" when is_nil(value) ->
+          {:cont, {:ok, acc}}
```
acco, line 603: great catch!

acco (approved):

acco:
@davoclavo note I also fixed the tests -- a lot of our database create paths actually connect to the db and check its configuration. So we need to reference test tables that we know are in the database (see test/support/unboxed_repo/migrations/20240816005458_create_test_tables.exs and an example model in test/support/models/character.ex)

So, I modified to refer to existing tables:
https://github.com/sequinstream/sequin/pull/1432/commits/935deb828aeca2f8f9f6b22173d32b9d2ceca0ae#diff-6c8c71c18120f4015b13a5e5560426e89afc5e883b7af3f33c1beb13fcaf7bca

PR #1441, opened 2025-05-12
fix: remove the brew tap command from CopyIcon for cli install

davoclavo:
Quick fix for the removal of `brew tap` from the CopyIcon cli installation instructions

cc @acco


acco:
Ah, thanks @davoclavo !

PR #1459, opened 2025-05-13
[minielixir] propagate line number in errors

### LINKED ISSUE #1452: Line number of compilation error of transformation function would be helpful
Can you render the line number of the error location when a transform function fails in the sink? If the transformation function is big (can have more than 100 lines) the user cannot figure out easily on which line the function broke 😶

Image

fazzone:
closes #1452 
Note that this PR only addresses errors in real life and NOT the transform editor/playground page.


lib/sequin/transforms/minielixir/validator.ex
```diff
@@ -25,11 +25,21 @@ defmodule Sequin.Transforms.MiniElixir.Validator do
...
         end
       end
     end
   end
 
+  def format_stacktrace(myself, stk) do
+    for {^myself, _f, _a, _info} = ste <- stk do
+      ste
+    end
+  end
```
RTLS, line 41: rm?

Issue #1359, opened 2025-05-05
Write a Jepsen-style test suite


acco:
**Background**<br>[Jepsen](https://jepsen.io/) is an open‑source framework for verifying the safety and liveness properties of distributed systems under adverse conditions such as network partitions, process crashes, and clock skew. Because Sequin streams database changes from Postgres to destinations like Kafka and SQS, we must be confident that optimizations or refactors to our pipeline do **not** violate correctness. A focused Jepsen test suite will give us a reproducible way to inject chaos and assert that Sequin still:

1. **Never loses a message** – every `INSERT`, `UPDATE`, and `DELETE` captured from Postgres is eventually delivered downstream. Duplicates are allowed but should be rare.
2. **Preserves user ordering** – for a given table, changes to the same primary‑key value must arrive in commit order (default is to group by primary-key(s)).

### Workload families we eventually want

| ID | Family | What it stresses |
| -- | -- | -- |
| **S‑TXN** | *Small transactions* | Many single‑row `INSERT`/`UPDATE`/`DELETE` ops per second across several tables. |
| **L‑TXN** | *Large multi‑row transactions* | Hundreds–thousands of rows changed atomically. (Interruptions happening in the middle.) |
| **B‑INT** | *Backfill + interactive traffic* | Ongoing backfill process in combination with S-TXN workloads. |
| **HI‑TPS** | *High‑throughput burst* | Sustained 5–10× normal throughput to uncover flow‑control issues. |

*Phase 1 implements only **S‑TXN**; the others follow in later phases.*

## Scope (Phase 1)

* Target a **single logical replication slot** replicating **multiple tables**.
* Implement the **S‑TXN** workload defined above.
* Leverage `Sequin.Havoc` (beginnings of a chaos module) for process‑level chaos; extend with network & sink faults where required.
* **Run locally** via (via `mix test` ? `mix task` ? Docker?)<br>*Future work:* add GitHub Actions CI job. And run in Docker.

## Chaos Events

| ID | Component | Fault | Injection path |
| -- | -- | -- | -- |
| **C1** | Connector processes (`slot_processor`, `sink_pipeline`, …) | Random `GenServer.stop/1` | `Sequin.Havoc` |
| **C2** | Network | Partition Postgres ⇆ Sequin for 15–60 s | `tc` / Linux `netem` |
| **C3** | Destination broker | Redis/Kafka/SQS downtime or connection interruption | stop & restart service |

## Implementation checklist

[...]

PR #1474, opened 2025-05-14
feat: add webhook integration test

abc3:
## About

Initial implementation for #1359 that validates transaction consistency through webhook events. The system performs multiple row updates in a table, generating update events that are received and validated for correct sequencing via webhooks.

## Running Tests

```bash
JEPSEN_TRANSACTIONS_COUNT="100" mix test --only jepsen
```

## Environment Variables

| Variable | Default | Description |
|----------|---------|-------------|
| `JEPSEN_HTTP_HOST` | "127.0.0.1" | HTTP server host |
| `JEPSEN_HTTP_PORT` | "4040" | HTTP server port |
| `JEPSEN_TRANSACTIONS_COUNT` | 20 | Number of transactions to run |
| `JEPSEN_TRANSACTION_QUERIES_COUNT` | 10 | Number of queries per transaction |

### DIFF
--- a/test/support/simple_http_server.ex
+++ b/test/support/simple_http_server.ex
@@ -8,7 +8,6 @@ defmodule Sequin.TestSupport.SimpleHttpServer do
 
   require Logger
 
-  @port Application.compile_env(:sequin, :webhook_test_port)
   @listen_options [:binary, packet: :raw, active: false, reuseaddr: true]
   @response """
   HTTP/1.1 204 No Content\r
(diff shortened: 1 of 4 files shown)


test/support/simple_http_server.ex
```diff
@@ -0,0 +1,58 @@
...
+  @moduledoc """
+  This module is a naive single-threaded HTTP server for webhook testing.
+  It accepts HTTP requests and forwards the parsed JSON body to the caller process.
+  Only intended for use in test environments.
+  """
+  use GenServer
+
+  require Logger
+
+  @port Application.compile_env(:sequin, :webhook_test_port, 8888)
```
abc3, line 11:
```suggestion
  @port Application.compile_env(:sequin, :webhook_test_port)
```

fazzone:
Looking good!  One thing I am curious about - maybe it's not worth it - what do you think about listening on port 0 and asking the socket what port you got assigned by the OS?  The HTTP endpoint can be created in setup with a nonsense port and then there would be a step in the test which does a genserver call to ask for the real port and updates the endpoint accordingly.  Or, the creation of the endpoint&sink could be done in a helper function rather than setup and be "private" to the individual test

abc3:
@fazzone This is a great suggestion, but I'm concerned it might be overengineering. It might be easier to handle this through some env var if more flexibility is needed

acco:
@abc3 Looks great! Will be very useful to build on this foundation. For example, I'd love to add erroring sinks to this (the SimpleHttpServer deterministically returns a 400)

It runs fast enough (2-3s) that I figured we can just have this in our normal test flow

PR #1476, opened 2025-05-14
fix: show missing Transform Function form validation errors

### LINKED ISSUE #1438: Fix crash when saving a transform with a duplicate/taken name
When trying to save a transform, if that transform's name is already taken, we do not gracefully show an error.

davoclavo:
Fix showing missing Function form validation errors that were not being shown because they were accessing the wrong `formErrors` accessor paths.

This caused subsequent DB-related constraint errors (returned by `Repo.insert`) such as "name has already been taken" to not show up, as those validations are not happening in the first place due to the failed initial validations.

This seems to be the expected default behavior as [stated in the docs](https://hexdocs.pm/ecto/Ecto.Changeset.html#module-validations-and-constraints) and even though I would love to see those DB-related errors as soon as possible, it may involve a deeper set of changes and implications.

Additional, unrelated change:
Always show the "Create Function" button on the same spot, to be consistent. Thus moving the Copy to GPT from the left side of that button to the right side.

Fixes #1438

### DIFF
--- a/assets/svelte/transforms/Edit.svelte
+++ b/assets/svelte/transforms/Edit.svelte
@@ -608,9 +608,9 @@ Please help me create or modify the Elixir function transform to achieve the des
               />
             </SelectContent>
           </Select>
-          {#if showErrors && formErrors.transform?.type}
+          {#if showErrors && formErrors.transform && Array.isArray(formErrors.transform)}
             <p class="text-sm text-red-500 dark:text-red-400">
-              {formErrors.transform.type[0]}
+              {formErrors.transform[0]}
             </p>
           {/if}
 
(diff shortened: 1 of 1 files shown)


fazzone (approved):
the error reporting seems good! As for the button, I'm certainly in favor of consistency but I'm not quite sure what you mean - was it that the buttons would move around depending on what transform type you selected?

davoclavo:
@fazzone yes exactly that! in general I think its good UX to keep the order/position of things consistent as much as possible; that said I might be biased to make my suggestion as I was changing them often for testing purposes.

happy to revert that if you think its good as it is tho

fazzone:
@davoclavo I agree completely! Just wanted to make sure that I was understanding everything.

PR #1493, opened 2025-05-15
fix: avoid saving form to local storage if submission was successful

### LINKED ISSUE #1472: DB connection form overlaps between different DBs causing it to error
Currently when I freshly open the dashboard and click on a db A first, try to edit it and call the "Test Connection" it works fine. But when I go back and click on another database, the previous db A form is not updated on the front end. But the db id is changed in the browser url. Causing it to error when I edit and update the database.

Looks like a bug.

davoclavo:
It seems that due to the use of `push_navigate` on a successful `form_submitted` event the callback receives `null` as a response, rather than `{ok: true}`

So now we just do the opposite check on submission: if the form response is `null` we don't restore to local storage, and if its `{ok: false}` we keep the form in local storage

Fixes #1472

Shoutout to @g14a for the amazing replication video! it was super helpful

--

Pre-fix replication steps:

1. Create two database records
2. Open up one, dont change anything, and click "Update Database" to cause a successful update
3. Open up the other db record, and you should see the polluted form details loaded from local storage

### DIFF
--- a/assets/svelte/databases/Form.svelte
+++ b/assets/svelte/databases/Form.svelte
@@ -185,7 +185,7 @@
     validating = true;
     clearFormStorage();
     pushEvent("form_submitted", { form }, (reply) => {
-      if (reply?.ok !== true) {
+      if (reply?.ok === false) {
         validating = false;
         progress.set(0);
         saveFormToStorage();


acco:
Heck yeah 🎉 🎢

PR #1494, opened 2025-05-15
fix: only show database exit confirm dialog when form is dirty

davoclavo:
Only show database exit confirm dialog when form is dirty, using the same implementation as the sink consumer form.

Also set the ssl default value to `true` in the `typed_schema` definition rather than in the `render` fn, as doing `database.ssl || true` always causes the value to be `true` regardless of the existing boolean set in `ssl`


acco:
This is great, thanks! 🎉 🥇

PR #1496, opened 2025-05-15
fix: update sequences inline on creating sink consumer to avoid race condition

### LINKED ISSUE #1437: Sometimes, replica identity alert message is missing table information
**Background**

When you first setup a sink, we kickoff a config check worker that see what the replica identity of the table is. If it's not `full`, we show an error message, as this means we won't receive `old` for changes, which is what most people want. Sometimes, that error message is wrong because it's missing the `schema` and `table` name:

**Fix**

**Step 1: Find out why this is happening.** It will not happen every time. I recommend trying to create a database then creating a sink right after and seeing what happens.

**Step 2: Fix the issue.**

davoclavo:
Updating db sequences for a sink consumer are performed asynchronously in the ConsumerLifecycleEventWorker.
This leads to a race condition where the sink details are shown before the sequences are updated, resulting in error messages that may be missing schema and table names.

Fixes #1437

### DIFF
--- a/lib/sequin/consumers/consumers.ex
+++ b/lib/sequin/consumers/consumers.ex
@@ -19,6 +19,7 @@ defmodule Sequin.Consumers do
   alias Sequin.Consumers.SinkConsumer
   alias Sequin.Consumers.SourceTable
   alias Sequin.Consumers.Transform
+  alias Sequin.Databases
   alias Sequin.Databases.PostgresDatabase
   alias Sequin.Databases.PostgresDatabaseTable
   alias Sequin.Databases.Sequence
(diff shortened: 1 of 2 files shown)


davoclavo:
Another option is to use the table information extracted and encoded [here](https://github.com/sequinstream/sequin/blob/main/lib/sequin_web/live/sink_consumers/show.ex#L623):

```elixir
  defp encode_consumer(%SinkConsumer{type: _} = consumer) do
    table = find_table_by_oid(consumer.sequence.table_oid, consumer.postgres_database.tables)
```

However I am unsure if that is a reliable way, as we currently rely on the table information being extracted from the consumer.sequence struct

Any feedback is welcome :)

acco:
Great find with this!

This is all good - we want to get rid of sequences ASAP, so I'm good with the current implementation

PR #1501, opened 2025-05-15
fix: unify consumer sink :new and :edit into a single form live view

### LINKED ISSUE #1497: UI Issue on top left when editing a sink
Image

Also the banner can be more descriptive like "Editing ${sink-name} Sink"

davoclavo:
Unify rendering Sink :new and :edit actions into a single unit of logic, in order to re-use as many components as we can, and also use the same layer without sidebar. Also refactor a couple of things to make things a bit more clear to edit/understand.

Also added the sink name to the title

Fixes #1497 

@g14a thanks for reporting this issue and suggesting improvements! Keep them coming 🙌🏼

### DIFF
--- a/lib/sequin_web/live/sink_consumers/form.ex
+++ b/lib/sequin_web/live/sink_consumers/form.ex
@@ -2,23 +2,12 @@ defmodule SequinWeb.SinkConsumersLive.Form do
   @moduledoc false
   use SequinWeb, :live_view
 
-  alias Sequin.Consumers.AzureEventHubSink
-  alias Sequin.Consumers.ElasticsearchSink
-  alias Sequin.Consumers.GcpPubsubSink
-  alias Sequin.Consumers.HttpPushSink
-  alias Sequin.Consumers.KafkaSink
-  alias Sequin.Consumers.NatsSink
-  alias Sequin.Consumers.RabbitMqSink
-  alias Sequin.Consumers.RedisStreamSink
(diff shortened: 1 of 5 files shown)


lib/sequin_web/live/sink_consumers/form.ex
```diff
@@ -30,200 +31,76 @@ defmodule SequinWeb.SinkConsumersLive.Form do
...
-    <.live_component
-      current_user={@current_user}
-      module={ConsumerForm}
-      id="new-consumer"
-      action={:new}
-      consumer={%SinkConsumer{type: :redis_stream, batch_size: 10, sink: %RedisStreamSink{}}}
-    />
-    """
+  defp fetch_or_build_consumer(socket, _kind, id) do
+    SequinWeb.SinkConsumersLive.Show.load_consumer(id, socket)
```
acco, line 62: I think we should just duplicate the load logic here (vs calling the LiveView) - we don't treat LiveViews as regular modules so don't call their pub functions

acco (changes requested):
Great refactor! See minor comment in-line

lib/sequin_web/live/sink_consumers/form.ex
davoclavo: On it!

davoclavo: Done! I'm glad you suggested this change because I realized a simpler query was needed without any preloads and health data

PR #1533, opened 2025-05-17
chore: replace Enum.map with Enum.each for side effects

abc3:
Replaced Enum.map with Enum.each since the mapped result wasn't used, to avoid unnecessary list allocation.


acco:
Thanks!

Issue #1534, opened 2025-05-17
Redis client doesn't process :eredis.qp result correctly, always returns :ok


abc3:
Since `:eredis.qp/2` returns a [list of values](https://github.com/Nordix/eredis/blob/502cd22b32dce6900f3d6b2746c808a269108ff4/src/eredis.erl#L160-L161), the result in the following case will always be `:ok`

https://github.com/sequinstream/sequin/blob/b33144d5868434f9d9e2e649693fc01821f66884/lib/sequin/sinks/redis/client.ex#L89-L92

acco:
cc @RTLS @fazzone

fazzone:
nice find!  we tested that this config worked, but. I guess we never saw any errors from redis.  Curious if there was a tool that surfaced this or just the mk1 eyeball

abc3:
@fazzone [credo](https://hexdocs.pm/credo/overview.html) and [dialyxir](https://hexdocs.pm/dialyxir/readme.html) can help. Added a PR #1547

PR #1547, opened 2025-05-19
feat: add credo and dialyxir for code quality checks

abc3:
This PR adds Credo and Dialyzer to the project and configures them in a github action

Update: Removed auto-check from the GA, as Dialyzer is [unhappy](https://github.com/sequinstream/sequin/actions/runs/15118561268/job/42495298685)

to run locally:
```elixir
mix dialyzer
```

```elixir
mix credo
```


acco:
@abc3 Thanks for this – looks like these are both setup to be optionally run, is that right?

**Dialyzer**
I don't think Dialyzer is worth the effort. We had "Dialyzer 0" on our last project, and I'm not convinced it was worth the ROI.

The investment:
- Time to run CI/get PRs approved
- Time invested to fix Dialyzer when it's broken in a developer environment
- Time getting nerd sniped by a Dialyzer error that a developer can't help but try to identify/fix

The return was it caught a handful of bugs. Lot of other higher yield time investments to catch bugs or improve the system!

FWIW I still have Dialyzer running in my IDE and it _can_ be helpful when building new stuff as part of the feedback loop. 

**Credo**
I think there's a new emerging lib that's like Styler but for Credo? Ideally, the tools fix the issues vs tell them they are there.

abc3:
@acco Yeah, they are run optionally. I added them because of [this comment](https://github.com/sequinstream/sequin/issues/1534#issuecomment-2891612393). Dialyzer detects an issue with :eredis.qp. However, feel free to reject this PR if you don't think it's worth it.

PR #1552, opened 2025-05-19
Add Kinesis sink

acco:
## Summary
- add Kinesis sink schema and AWS helper
- add Kinesis pipeline implementation
- wire Kinesis into consumer and runtime
- support Kinesis in factories and YAML loader
- add basic tests for AWS client and pipeline
- add frontend support for the Kinesis sink


lib/sequin/runtime/kinesis_pipeline.ex
```diff
@@ -0,0 +1,55 @@
...
+
+    setup_allowances(test_pid)
+
+    records =
+      Enum.map(messages, fn %{data: data} ->
+        %{
+          "Data" => Base.encode64(Jason.encode!(Sequin.Transforms.Message.to_external(consumer, data))),
+          "PartitionKey" => data.group_id
+        }
+      end)
```
acco, line 41: dedupe field/idempotency key we can use?

fazzone (approved):
🐑

PR #1587, opened 2025-05-22
refactor: remove legacy_event_singleton_transform

abc3:


test/sequin/http_push_pipeline_test.exs
```diff
@@ -139,33 +139,6 @@ defmodule Sequin.Runtime.HttpPushPipelineTest do
       assert_receive :sent, 1_000
     end
 
-    test "legacy_event_singleton_transform sends unwrapped single messages", %{
```
abc3, line 142: is it right to remove it, or should we keep this test?

PR #1614, opened 2025-05-23
Add constraint to annotations (avoid overload)

acco:


RTLS (commented):
Will this raise when Ecto runs an insert or update operation with annotations that violate this payload size? should we convert those to an error instead of raising?

After a quick glance through the Changeset docs I'm actually not sure how to convert a database constraint to an error, other than foreign keys and unique constraints.

Perhaps just:

```
  defp validate_annotations_size(changeset) do
    validate_change(changeset, :annotations, fn :annotations, annotations ->
      # Convert the map to JSON string to measure byte size
      json_string = Jason.encode!(annotations)
      byte_size = byte_size(json_string)
      
      if byte_size <= @max_annotation_size_bytes do
        []
      else
        [annotations: "is too large (#{byte_size} bytes, maximum is #{@max_annotation_size_bytes} bytes)"]
      end
    end)
```

acco:
Good call @RTLS added

PR #1624, opened 2025-05-26
feat: persist pagination state when navigating to/from sink details

### LINKED ISSUE #1618: Use query params to save the page on /sinks
If you navigate into a sink then back out to the /sinks page, we'll lose your page. We can change this by storing the page as query params in the URL.

davoclavo:
- Update sink consumers index LiveView to handle page query parameter in mount and handle_params
- Add pageStore to maintain current page number across navigation between Index and Show sink components
- Update sink Index to store current page before navigating to details
- Modify back button in Show sink to return to correct page

Fixes #1618

### DIFF
--- a/assets/svelte/consumers/ShowSinkHeader.svelte
+++ b/assets/svelte/consumers/ShowSinkHeader.svelte
@@ -1,5 +1,6 @@
 <script lang="ts">
   import { onMount } from "svelte";
+  import { pageStore } from "../stores/pageStore";
   import {
     ArrowLeft,
     Clock,
@@ -133,7 +134,9 @@
   <div class="container mx-auto px-4 py-4">
     <div class="flex items-center justify-between">
       <div class="flex items-center space-x-4">
-        <LinkPushNavigate href="/sinks">
+        <LinkPushNavigate
+          href={$pageStore ? `/sinks?page=${$pageStore}` : "/sinks"}
+        >
           <Button variant="ghost" size="sm">
             <ArrowLeft class="h-4 w-4" />
           </Button>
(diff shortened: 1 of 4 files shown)


acco (approved):
@davoclavo Looks great! So we can use history if the user hits the back button, but not the little back arrow in the top left – is that right?

davoclavo:
@acco Ahhh great question! I hadn't thought about using `history.back()` until now, so I did a bit of digging and here is what I found:

Using `history.back()` generally works fine if the user navigated through the app normally. But there is at least one edge case: if you land on the sink page directly (e.g. if someone shares a link, or you pasted it in the browser address bar). In that case there are two scenarios:

1. History is empty - (which we can detect using `history.length`) - we can perform a normal live redirect to `/sinks`

2. History is not empty - in this case we run into troubles as we don't know if `history.back()` will take the user to the last visited sinks page, or to some other external website that they were visiting earlier. I guess for security reasons you cannot inspect the history stack, and I couldn't find or think of a reliable workaround using browser APIs (such as `window.referrer`).

So yes, I would suggest to rely on the svelte store implementation as it gives us a predictable way to control navigation within the app. That said, I am no expert on this, and perhaps there are other better approaches that we could explore!

Did I understand your question properly tho? I hope so 😅 

---

On a side note, I am also thinking about modifying the `change_page` event handler, to only keep the latest sink page on the history stack, rather than every single sink page. Any thoughts?

```diff
   def handle_event("change_page", %{"page" => page}, socket) do
     account_id = current_account_id(socket)
     total_count = Consumers.count_sink_consumers_for_account(account_id)

     socket =
       socket
       |> assign(:page, page)
       |> assign(:total_count, total_count)
       |> assign(:encoded_consumers, nil)
       |> async_assign_consumers()
-      |> push_patch(to: ~p"/sinks?page=#{page}")
+      |> push_navigate(to: ~p"/sinks?page=#{page}", replace: true)
```

acco:
All sounds good, and I like change_page! Merging

PR #1628, opened 2025-05-27
fix UI issue where a long csv of filters expands and doesn't wrap

g14a:
UI expands and doesn't give a good feel when there's a long list of a column filter in CSV fashion

Before:

image

After:

image


acco:
Much appreciated, thanks @g14a !

PR #1662, opened 2025-05-29
feat: persist function editable code and test message fields in local storage

### LINKED ISSUE #1601: For transform functions, save draft in localStorage
so that even if the server crashes or restarts you can't lose work

davoclavo:
This PR adds local storage persistence for edited function test message fields (both for synthetic and real messages)

- Changes are cleaned up when form is closed or saved
- Adjusted synthetic and test message IDs to have fixed/deterministic UUID per message to keep track of the field changes _per message_

Misc improvements:
- Moved action/field types to types.ts and added proper TypeScript enums

Fixes #1601

-----

### Demo

NOTE: When the screen scrolls to the top after I save editions it means that I refreshed the browser to simulate a disconnection or just a page reload.

https://github.com/user-attachments/assets/1f704350-f0e0-46b8-98a6-d1c9d777fad4


lib/sequin/runtime/message_handler.ex
```diff
@@ -203,6 +203,7 @@ defmodule Sequin.Runtime.MessageHandler do
     payload_size = :erlang.external_size(data)
 
     %ConsumerEvent{
+      id: message.trace_id,
```
davoclavo, line 206: @RTLS I noticed this `id` field was [removed recently](https://github.com/sequinstream/sequin/commit/c43295b9146049e5143b36e1bd226625d4de28bd) - was it causing issues? (can't really tell as all tests are passing)

If the answer is yes: I need to figure out a way to use a unique and fixed value per message to keep track of the edited state in the UI

If the answer is no: Is it ok if I use this field? or would you recommend me try to figure out an alternative?

fazzone:
This looks good so far and it's a great change but the original idea of my issue #1601 was to save the code of the function itself.  If you feel that is easy to do in this PR then I recommend it! A new PR for only that, either before or after this one, is fine too if you prefer.

lib/sequin/runtime/message_handler.ex
davoclavo: Thanks for the context, I'll try to use the trace_id

davoclavo:
@fazzone ohhh makes sense. Let me take a look into what would it take to persist the transform function body, and if its trivial I'll append it to this PR. Thanks for letting me know!

davoclavo:
@fazzone I just adjusted my implementation to use `replication_message_trace_id` instead of the `id` attribute to map changes between test messages. Thanks for your suggestion!

Regarding persisting the function code itself, I will make those changes in a separate PR. So feel free to merge this one whenever you deem appropriate.

davoclavo:
Changed my mind at the last minute and included persisting function code here as well! 😅

fazzone:
Okay, great!  I want to do some local testing of this before merging so it may be a little while.  Talking with @RTLS , another candidate ID you can use is the tuple `consumer_id,commit_lsn,commit_idx` The only potential issue with the trace ID, which I didn't realize at the time, is that it may be affected by a system restart.  I actually don't think that is a blocker at all because I believe the worst thing that would happen is that you might see duplicates?  Just including that for full context

davoclavo:
@fazzone thanks for the context around the trace_id, let me test it restarting the server and see if it causes issues. I could use the tuple, but will be a bit verbose unless I make a composite I think ([here are the changes I had to make from `id` -> `replication_message_trace_id`](https://github.com/sequinstream/sequin/pull/1662/commits/5047edd698d0a01e13c75df33feec2627784fa17)

fazzone:
Yes I think the ideal way to use the tuple would be to just have a function which computes it with string concat and then use that wherever possible and maybe store it as _key or or something on the frontend objects?  Easier said than done of course!  Like I said, I don't think it's important, I expect the trace ID should be perfectly fine

davoclavo:
I think using the trace_id will work for now, because the synthetic message will always have a static trace_id, and for real test messages if the server is restarted I think the ets table storing them gets cleared anyways, so new messages would have to arrive.

fazzone:
Makes sense, sounds good!

davoclavo:
@fazzone Howdy! I just added a couple of improvements based on your suggestions. Let me know what you think :)

List of changes:

- We also now store function type, and sink type in local storage, so users don't have to select it again
- Support storing in local storage for **editing** functions, not just new ones. (I initially thought functions were immutable, so that's why I skipped it in the first place)
- Add a button to the bottom right of the editor to restore to the original code, and a label to indicate that code has been modified, so users can tell whether the code being shown is actually persisted in the DB or loaded from the local storage
  - For new functions: this resets to the initial static code for the selected function type+sink
  - For editing functions: this resets to the persisted code in the backend
  
  
---

image

fazzone:
Nice, it looks good.  I haven't had the time to check out locally and experiment but I have a couple points:
- It seems that you can have multiple independent drafts saved in localstorage depending on type/sink_type.  So the expected experience in case of a backend crash would be that you have to remember to select the same type/sink_type you were using before, and then your code would be populated?  I think this is not unreasonable, but it's potentially not the most obvious - I feel like as a user experiencing a backend crash, I would still be shocked and annoyed that the computer ate my precious input when everything gets cleared out and I get shown an empty form.  While it's true that this frustration would only persist for a moment before being resolved, I would argue it's a better user experience for it to never happen in the first place.  
- That doesn't mean that one-draft-to-rule-them-all is the obvious right way to go - it just means that it should be balanced against the usefulness of having multiple drafts.  However, right now I feel as though I don't really understand why multiple drafts is useful, which pretty much explains my preference for one draft.  Just curious how you are thinking about that
- Despite the fact that i absolutely LOVE the little trash can icon, I would actually personally propose a complete ban on the use of the trash can icon in software that is designed to connect to production databases.  I would consider instead a circling-backwards-arrow, which I think was used as the "cancel" for the editable test data fields?  Note that when we have gotten all the way to bikeshedding icons it means everything else looks good!

fazzone:
Okay, that makes sense!  I didn't read carefully enough to fully understand the behavior.  Really appreciate the detailed explanation as well as the video demos.  I am now convinced that it makes sense, especially by the example of editing multiple functions at once, which I hadn't considered.

assets/svelte/functions/functionLocalStorage.ts
```diff
@@ -0,0 +1,38 @@
+import type { FormData } from "./types";
```
RTLS: functionCodeStorage

lib/sequin_web/live/functions/edit.ex
```diff
@@ -274,10 +274,10 @@ defmodule SequinWeb.FunctionsLive.Edit do
       Map.put(
         form_errors,
         :modified_test_messages,
-        Enum.reduce(modified_test_messages, %{}, fn {id, result}, acc ->
+        Enum.reduce(modified_test_messages, %{}, fn {replication_message_trace_id, result}, acc ->
```
RTLS, line 277: idempotency_key

RTLS (approved):

PR #1684, opened 2025-05-30
feat: make actual links for sink index table so users can cmd/ctrl+click

### LINKED ISSUE #1670: Make sink rows on the Sink#Index page actual links
You can't CMD+Click to open sinks on our Index page. We can support that by either:

* Making rows links
* Having a separate button you click to open sinks

We should look at comps from other similar products

davoclavo:
Make each row of the Sink Index page a link so users can cmd/ctrl+click on them to open in new tabs.

Also fix LinkPatchNavigate component to scroll to top as users might be scrolled all the way to the bottom and clicking on a sink should take them to the top of the page.

Fixes #1670

### DIFF
--- a/assets/svelte/components/LinkPatchNavigate.svelte
+++ b/assets/svelte/components/LinkPatchNavigate.svelte
@@ -3,7 +3,13 @@
   export let href;
 </script>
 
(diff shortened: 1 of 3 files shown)


acco (approved):
Looks great, thanks @davoclavo !

PR #1689, opened 2025-05-30
💄 Pretty trace empty state when unpaused

RTLS:
* Also clean up handle_info in show.ex for trace events


RTLS:
https://linear.app/sequin/issue/SQN-3644/trace-tweak

PR #1691, opened 2025-05-30
[cli] allow function code to be embedded via `file:` + smarter ENV embed in yaml

### LINKED ISSUE #1584: Support loading files into sequin.yaml
Would love to be able to:

```
transforms:
  - name: typesense-transformer
    type: function
    file: sequinTransformer.exs
```

fazzone:
Previously, the cli did not parse yaml and applied env var subst to the entire yaml as a flat string.  
This PR adds a real yaml parser and applies env var substitution in a structure-aware way, so that you can't interpolate into yaml keys, or interpolate in content which would result in malformed yaml.
Finally, with the parser, we can support a file: key on functions, where we will read the file, path relative to our cwd, and pass the content upstream under the code: key as the api expects.

Consider:
- should we keep the file: separate key, or do in-band signaling where we automatically read files into anything that starts with file:// or something?
- should we interpret paths as relative to process cwd, or relative to the location of the yaml file, or both?

closes #1584

### DIFF
--- a/cli/cli/config_command_test.go
+++ b/cli/cli/config_command_test.go
@@ -187,11 +187,11 @@ functions:
 		}
 
 		testCases := []struct {
-			name        string
-			filePath    string // Path to use in YAML
-			codePath    string // Actual path where code should be written
-			cleanup     func()
-			useStdin    bool   // Whether to use STDIN for YAML input
+			name     string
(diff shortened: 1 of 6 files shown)


RTLS (approved):
* add relative search path so .exs files can be relative to .yaml

PR #1696, opened 2025-05-30
feat: add Oban.Plugins.Pruner to delete completed jobs older than 14 days

### LINKED ISSUE #1687: Add an Oban cron job that cleans up the `oban_jobs` table
We should have an Oban cron job (enqueued [via cron config](https://github.com/sequinstream/sequin/blob/f46320ad03268f101ebc341949f16a6b292d2994/config/config.exs#L36-L66)) that cleans the `oban_jobs` table. It should run every 8 hours. It should delete entries that are older than 14 days.

It should try to clean the table as quietly as possible. One good strategy for that is to delete only 5k rows, with a 20s pause/interval between delete queries. It should run until the `num_rows` affected falls below 5k (or whatever the delete batch is).

We can run those deletes all inside one job that just `Process.sleep`s between intervals--if Oban will let the job run for long enough. Alternatively, we can run one job that then enqueues its successor to run 20s later.

No matter what, take a look at other strategies we use to ensure duplicate jobs are not enqueued (if for some reason the job is running for too long, we don't accidentally start the next chain). e.g.:

[https://github.com/sequinstream/sequin/blob/f46320ad03268f101ebc341949f16a6b292d2994/lib/sequin/health/check_postgres_replication_slot_worker.ex#L4-L11](https://github.com/sequinstream/sequin/blob/f46320ad03268f101ebc341949f16a6b292d2994/lib/sequin/health/check_postgres_replication_slot_worker.ex#L4-L11)

davoclavo:
Add Oban Plugin [`Oban.Plugins.Pruner`](https://hexdocs.pm/oban/Oban.Plugins.Pruner.html) to clean up completed jobs older than 14 days.

Fixes #1687

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -44,6 +44,8 @@ config :sequin, Oban,
   ],
   repo: Sequin.Repo,
   plugins: [
+    # Prune jobs that are older than 14 days
+    {Oban.Plugins.Pruner, max_age: 14 * 24 * 60 * 60, limit: 5_000, interval: 20_000},
     {Oban.Plugins.Cron,
      crontab: [
        # Runs every 6 hours (at minute 0)


acco:
@davoclavo Great work finding this, thanks for not re-inventing the wheel!

PR #1711, opened 2025-06-02
feat: implement kafka end-to-end tests

abc3:
# E2E Tests Implementation

This PR adds end-to-end tests to validate the integration between PostgreSQL and Kafka using a demo table setup, along with automated daily testing via GitHub Actions.

### E2E Test Suite
- Implements tests to verify data streaming from PostgreSQL to Kafka
- Uses a demo table setup with auto-incrementing ID, text field, and timestamp
- Supports configurable test data volume via `TEST_MESSAGES_COUNT` env var
- Includes parallel data insertion for more realistic load testing
- Health checks for both Sequin and Kafka services

### Test Infrastructure
- Docker Compose setup with:
  - PostgreSQL database
  - Kafka broker
  - Redis for caching
  - Sequin service
- Make commands for easy test management:
  - `e2e-up`: Start containers
  - `e2e-down`: Clean up
  - `e2e-rebuild`: Fresh rebuild
  - `e2e-tests`: Run tests

### GitHub Actions Workflow
- Automated daily runs at 17:00 UTC (10:00 PDT)
- Runs E2E tests with fresh container builds
- Includes health checks for services
- Supports manual triggering via workflow_dispatch

## Testing
You can run the tests locally using:
```bash
make e2e-rebuild  # Fresh setup
make e2e-tests    # Run tests
```

For different data volumes:
```bash
TEST_MESSAGES_COUNT=500 make e2e-tests    # Small dataset
TEST_MESSAGES_COUNT=10000 make e2e-tests  # Large dataset
```


priv/tests_e2e/tests.exs
```diff
@@ -0,0 +1,125 @@
...
+    password: "postgres",
+    database: "sequin_test"
+  ]
+
+  @test_topics ["demo-topic"]
+
+  @expected_count String.to_integer(System.get_env("TEST_MESSAGES_COUNT", "1000"))
+
+  setup do
+    Process.sleep(5_000)
```
acco: Obviously, would be great to have more determinism here. How can we know the test suite is ready to execute?

priv/tests_e2e/tests.exs
```diff
@@ -0,0 +1,125 @@
...
+
+  defp insert_test_data(expected_count) do
+    {:ok, pid} = pg_conn()
+
+    for n <- 0..(expected_count - 1) do
+      {:ok, _} = Postgrex.query(pid, "INSERT INTO demo_table (demo_text) VALUES ($1)", ["Generated text entry ##{n}"])
+    end
+
+    GenServer.stop(pid)
+  end
```
acco: Would be great to do inserts in parallel (far closer to a production/real-world simulation)

priv/tests_e2e/tests.exs
abc3: we can determine the test suite is ready to execute by pinging the services when it starts

Makefile
```diff
@@ -134,3 +134,28 @@ docs: ## Run mintlify dev server for documentation
 redis-console-consumer: ## Read from redis stream <stream-key> [from-beginning]
 	@./scripts/redis-console-consumer.sh $(wordlist 2,$(words $(MAKECMDGOALS)),$(MAKECMDGOALS))
 
+e2e-up:
```
acco: add a make command to trigger `gh` action run too? By default for the current branch, maybe also with an optional `BRANCH` env/arg to target another branch?

priv/tests_e2e/docker-compose.yml
```diff
@@ -0,0 +1,122 @@
+version: '3.8'
+
+services:
+  postgres:
+    build: 
+      context: ../../
+      dockerfile: Dockerfile.postgres
+    command: postgres -c shared_preload_libraries=pg_stat_statements -c pg_stat_statements.track=all -c max_connections=200 -c pg_stat_statements.max=10000 -c track_activity_query_size=2048 -c wal_level=logical
+    ports:
+      - "127.0.0.1:5432:5432"
```
acco: can we have Postgres on another port so it doesn't conflict with any other postgres running on local machine? For example, I always have my `sequin_dev` Postgres running on `5432`

priv/tests_e2e/docker-compose.yml
```diff
@@ -0,0 +1,122 @@
...
+      test: ["CMD-SHELL", "pg_isready -U postgres"]
+      interval: 5s
+      timeout: 5s
+      retries: 10
+      start_period: 10s
+
+  redis:
+    image: redis:7
+    ports:
+      - "127.0.0.1:6379:6379"
```
acco: Same with Redis (different port)

priv/tests_e2e/tests.exs
```diff
@@ -0,0 +1,265 @@
...
+
+    on_exit(fn ->
+      GenServer.stop(conn)
+    end)
+
+    {:ok, %{conn: conn, tid: tid}}
+  end
+
+  setup %{conn: _conn, tid: tid} do
+    # clear_test_data(conn)
```
acco: just noting this, not sure if it should be commented out

priv/tests_e2e/tests.exs
```diff
@@ -0,0 +1,265 @@
...
+
+  defp assert_messages(messages, expected_count, tid) do
+    assert length(messages) >= expected_count,
+           "Expected #{expected_count} messages, but got #{length(messages)}"
+
+    Enum.reduce(messages, 0, fn message, acc ->
+      decoded_value = Jason.decode!(message.value)
+      assert decoded_value["action"] == "insert"
+      record = decoded_value["record"]
+      assert [{record["demo_text"]}] == :ets.take(tid, record["demo_text"])
```
acco, line 83: This pattern match will break if Sequin delivers a duplicate, is that right?

We think that duplicates should be rare in this situation. Sequin should only deliver duplicates in strange crash situations where we write to the destination but fail to write to Redis before crashing.

We're thinking it's OK to start super strict and basically assert that messages == expected_count and that there are no duplicates in the destination! In the future, we can always relax this -- we'd relax the duplication check in a situation where we were crashing parts of Sequin in the test.

abc3: made an ets duplicate_bag to catch duplicates

acco:
Great work here, thanks @abc3 !

PR #1727, opened 2025-06-04
fix: improve secrets yaml serialization

### LINKED ISSUE #1442: cli bug: exporting and loading http_endpoint with encrypted headers with `--no-show-sensitive` not working properly
Exporting and importing `http_endpoints` with encrypted headers via the CLI with `--no-show-sensitive` does not work properly.

Steps to replicate:

1. Create an `http_endpoint` **with** an encrypted header
2. Export sequin config to a yaml with `sequin config export --no-show-sensitive >| sequin.yaml`
3. Try to make a config apply diff with `sequin config plan sequin.yaml`
4. Fails with the following error:

```
sequin: error: server returned error: {"summary":"Invalid HTTP endpoint configuration: Error updating HTTP endpoint 'fragrant_water_42': Invalid headers configuration. Must be a list of key-value pairs.\n\nHTTP Endpoints are destinations for Webhook Subscriptions.\n\nThey can be configured in one of three ways:\n\n1. Configure an internet accessible URL:\n\nhttp_endpoints:\n  - name: \"external-endpoint\"\n    url: \"https://api.example.com/webhook\"\n\n2. Configure a local endpoint that uses the Sequin CLI to create a secure tunnel:\n\nhttp_endpoints:\n  - name: \"local-endpoint\"\n    local: \"true\"\n    path: \"/webhook\"\n\n3. Configure a Webhook.site endpoint for development purposes:\n\nhttp_endpoints:\n  - name: \"webhook.site-endpoint\"\n    webhook.site: \"true\"\n\n\nShared options:\n\n- name: \"some-name\"\n- headers:\n    - key: \"X-Header\"\n      value: \"my-value\"\n- encrypted_headers:\n    - key: \"X-Secret-Header\"\n      value: \"super-secret\"\n\n"}
```

`sequin.yaml` fragment:

```yaml
...

http_endpoints:
  - name: bold_sound_55
    url: http://localhost:4000/something
    headers: {}
    encrypted_headers: {}
  - name: fragrant_water_42
    url: http://localhost:4000/testing
    headers: {}
    encrypted_headers: '(1 encrypted header(s)) - sha256sum: cf529592'

...
```

davoclavo:
Improve serialization of encrypted headers and gcp pubsub credentials.

- Instead of exporting encrypted_headers and gcp pubsub credentials as a string with the sha256 we now export the map structure with obfuscated values
- Also fix exporting headers and encrypted headers for webhook site http_endpoints

Fixes #1442


lib/sequin/transforms/transforms.ex
```diff
@@ -449,20 +466,8 @@ defmodule Sequin.Transforms do
...
-  defp sha256sum(encrypted_headers) do
-    encrypted_headers
-    |> :erlang.term_to_binary()
-    |> then(&:crypto.hash(:sha256, &1))
-    |> Base.encode16(case: :lower)
-    |> String.slice(0, 8)
+    Map.new(encrypted_headers, fn {key, value} -> {key, maybe_obfuscate(value, false)} end)
   end
 
   defp maybe_obfuscate(nil, _), do: nil
```
acco, line 473: Unrelated to your changes, but we should use `Sequin.String.obfuscate` for the last clause here!

acco (approved):
Looks great! One minor change, then 🚢

lib/sequin/transforms/transforms.ex
davoclavo: @acco Thanks for the feedback! I initially opted to not use `Sequin.String.obfuscate` to be consistent with how other secrets are being obfuscated in the YAML export.

In the case you think we should use that function, I would recommend to change the underlying `maybe_obfuscate` implementation to use that function instead of hardcoding `'******'` for all cases of the export. ([source](https://github.com/sequinstream/sequin/blob/main/lib/sequin/transforms/transforms.ex#L470))

What do you think? I am also happy to just do that for this particular case, but I believe that consistency can be good.

acco: @davoclavo Yeah, that's what I think/agreed!

davoclavo: Done ✅

PR #1736, opened 2025-06-04
feat: improve plan output for sensitive values

### LINKED ISSUE #1298: Improve handling of secrets in `sequin.yaml`

davoclavo:
## Improve sensitive value handling in config plan diffs

This PR addresses how sensitive values are displayed during config plan diffs. Previously, sensitive fields could be omitted entirely from diffs if their obfuscated values matched.

### Changes
- Introduced `SensitiveValue` struct to represent sensitive values without immediate obfuscation. Also added corresponding implementations to encode them as strings, json, and yaml.
- Updated `Map.reject_nil_values` to handle `SensitiveValue` structs

### Testing
- Added a tests to show how diffs are rendered sensitive value handling
- Modified existing tests cases to reflect usage of the new `SensitiveValue` struct

Any feedback is more than welcome!

Fixes #1298

----

## New diffs
image

### DIFF
--- a/lib/sequin/map.ex
+++ b/lib/sequin/map.ex
@@ -42,13 +42,7 @@ defmodule Sequin.Map do
 
   def reject_nil_values(map) do
     map
-    |> Enum.reject(fn {_, v} ->
-      case v do
-        %Sequin.Transforms.SensitiveValue{value: nil} -> true
-        nil -> true
-        _ -> false
-      end
-    end)
+    |> Enum.reject(fn {_, v} -> is_nil(v) end)
     |> Map.new()
   end
 
--- a/lib/sequin/transforms/transforms.ex
+++ b/lib/sequin/transforms/transforms.ex
@@ -251,7 +251,7 @@ defmodule Sequin.Transforms do
   end
 
   def to_external(%KafkaSink{} = sink, show_sensitive) do
-    Sequin.Map.reject_nil_values(%{
+    reject_nil_values(%{
       type: "kafka",
       hosts: sink.hosts,
       topic: sink.topic,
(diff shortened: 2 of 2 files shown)


lib/sequin/map.ex
```diff
@@ -42,7 +42,13 @@ defmodule Sequin.Map do
 
   def reject_nil_values(map) do
     map
-    |> Enum.reject(fn {_, v} -> is_nil(v) end)
+    |> Enum.reject(fn {_, v} ->
+      case v do
+        %Sequin.Transforms.SensitiveValue{value: nil} -> true
```
acco: nit: We want `Sequin.Map` et al to behave like Elixir std lib extensions, so they shouldn't know anything about our codebase. Makes them very unsurprising to use.

I'd instead just write a new `reject_nil_values` private function in Transforms that does this / treats SensitiveValue as nil

PR #1749, opened 2025-06-06
fix: close postgrex connections and stop replication on database deletion

abc3:
When a database is deleted, this stops the Postgrex connection and replication slot. Previously, Sequin left them running, which caused zombie processes


abc3:
need to fix tests

acco:
@abc3 I suspect that pid/conn here might sometimes be a conn from our Sequin.Repo test connection pool, hence the failures after we kill it?

We could potentially move this call to [DatabaseLifecycleEventWorker](https://github.com/sequinstream/sequin/blob/929ab4c68ff08bdeea3773a99d409246f1b0a53e/lib/sequin/runtime/database_lifecycle_event_worker.ex#L1) -- could be a good spot for it. Not sure if that would fix the problem.

PR #1787, opened 2025-06-10
fix: only ignore esc keypress for the function code editor

### LINKED ISSUE #1781: Do not close function form on `escape` when editing
We use CodeMirror for editing functions (transforms/filters). it's common when using a text editor to use `escape` to try to e.g. close auto-complete/intellisense. However, that tries to close the whole form!

Let's disable escape for closing when the cursor is in the function editor?

davoclavo:
Fixes #1781 specifically for the function code editor, and not for the whole FullPageForm component as done by https://github.com/sequinstream/sequin/pull/1783


acco:
You're the man, @davoclavo !

Issue #937, opened 2025-01-31
Alert/notify when a message is rejected due to size


acco:
A message may be rejected due to size:

* There's a max message size of `5mb` in the SlotProcessor (currently)
* Some downstream sinks have max message sizes of their own (e.g. SQS)

We should log somewhere when these messages are rejected for users to review

RTLS:
Good call out on ingestions vs. delivery failures due to max message size

These should probably be handled with different errors. For Sequin, we don't want to ingest at all. For downstream, we may want to treat this message as a normally failed message because some sinks have configurable max message sizes and users can re-configure their sink

PR #1795, opened 2025-06-11
Add S2 sink

acco:
## Summary
- add S2Sink schema and client with connection test
- integrate S2 sink into consumer form and pipelines
- add S2 frontend components and types
- update logger redactor for S2 access token
- add factories and tests for S2 sink and pipeline

## Testing
- No tests run due to environment limitations

------
https://chatgpt.com/codex/tasks/task_e_6849dc3340b08326bb8cab97ddc0c61d


lib/sequin/runtime/s2_pipeline.ex
```diff
@@ -0,0 +1,54 @@
...
+
+  @impl SinkPipeline
+  def handle_batch(:default, messages, _batch_info, context) do
+    %{consumer: %SinkConsumer{sink: %S2Sink{} = sink} = consumer, test_pid: test_pid} = context
+
+    setup_allowances(test_pid)
+
+    records =
+      Enum.map(messages, fn %{data: data} ->
+        %{"body" => Jason.encode!(Message.to_external(consumer, data))}
```
acco, line 40: ?

test/sequin/s2_client_test.exs
```diff
@@ -0,0 +1,33 @@
...
+  alias Sequin.Factory.SinkFactory
+  alias Sequin.Consumers.S2Sink
+
+  @sink %S2Sink{type: :s2, endpoint_url: "https://s2.dev", stream: "test-stream", access_token: "token"}
+
+  describe "append_records/2" do
+    test "successfully sends batch" do
+      records = [SinkFactory.s2_record(), SinkFactory.s2_record()]
+
+      Req.Test.stub(Client, fn conn ->
```
acco: expect

lib/sequin/consumers/s2_sink.ex
```diff
@@ -0,0 +1,37 @@
...
+  import Ecto.Changeset
+
+  alias Sequin.Encrypted
+
+  @derive {Jason.Encoder, only: [:endpoint_url, :stream]}
+  @derive {Inspect, except: [:access_token]}
+  @primary_key false
+  typed_embedded_schema do
+    field :type, Ecto.Enum, values: [:s2], default: :s2
+    field :endpoint_url, :string
```
davoclavo: Note:
Perhaps we should consider using `:basin` instead of `:endpoint_url` as I think the endpoint is derived from the basin+stream via `https://{basin}.b.aws.s2.dev/v1/streams/{stream}/records`

https://s2.dev/docs/rest/records/append-records

yordis: Agree, add a `basin` configuration until it is required to overwrite the entire endpoint for unknown reasons

assets/svelte/sinks/s2/S2Icon.svelte
```diff
@@ -1,35 +1,18 @@
...
-        id="Amazon-Kinesis-Data-Streams_Icon_64_Squid"
-      ></path>
-    </g>
+  <g id="s2">
```
yordis: be careful with `id`, they are unique across the entire web page, when in doubt, remove them unless they are necessary, or use something that wouldn't conflict with something else

assets/svelte/sinks/s2/S2SinkCard.svelte
```diff
@@ -6,7 +6,7 @@
 
   export let consumer: S2Consumer;
 
-  let s2Url = consumer.sink.endpoint_url;
+  let s2Url = `https://s2.dev/dashboard/basins/${consumer.sink.basin}`;
```
yordis: should this be passed from the backend `export let s3Url`, I would expect only one component to construct this URL to avoid problems

davoclavo: Good point. Currently the url that is generated by the backend is for the [REST API](https://github.com/sequinstream/sequin/blob/codex/add-s2.dev-sink/lib/sequin/consumers/s2_sink.ex#L20-L22) - but we could also add some url generators for the UI dashboard, such as this one

assets/svelte/sinks/s2/S2Icon.svelte
davoclavo: Good catch! fixed

assets/svelte/sinks/s2/S2SinkCard.svelte
davoclavo: Hmmm, after digging a bit more, it seems we would need to add that attribute to the `type_embedded_schema` (plus a couple of related changes) in order for it to be serialized for the front end.

I think I am more inclined to keep this in the UI for now, as we dont use it at all for any process in the backend, but if you think its worth it I'm down to commit these changes:

```diff
diff --git a/assets/svelte/consumers/types.ts b/assets/svelte/consumers/types.ts
index de1fe39e..9abc6844 100644
--- a/assets/svelte/consumers/types.ts
+++ b/assets/svelte/consumers/types.ts
@@ -109,6 +109,7 @@ export type S2Consumer = BaseConsumer & {
     basin: string;
     stream: string;
     access_token: string;
+    dashboard_url: string;
   };
 };

diff --git a/assets/svelte/sinks/s2/S2SinkCard.svelte b/assets/svelte/sinks/s2/S2SinkCard.svelte
index 025c5f7d..3b0a7268 100644
--- a/assets/svelte/sinks/s2/S2SinkCard.svelte
+++ b/assets/svelte/sinks/s2/S2SinkCard.svelte
@@ -5,8 +5,6 @@
   import type { S2Consumer } from "../../consumers/types";

   export let consumer: S2Consumer;
-
-  let s2Url = `https://s2.dev/dashboard/basins/${consumer.sink.basin}`;
 </script>

 <Card>
@@ -14,7 +12,11 @@
     <div class="flex justify-between items-center mb-4">
       <h2 class="text-lg font-semibold">S2 Configuration</h2>
       <div class="flex space-x-2">
-        <a href={s2Url} target="_blank" rel="noopener noreferrer">
+        <a
+          href={consumer.sink.dashboard_url}
+          target="_blank"
+          rel="noopener noreferrer"
+        >
           <Button variant="outline" size="sm">
             <ExternalLink class="h-4 w-4 mr-2" />
             View Basin in S2
diff --git a/lib/sequin/consumers/s2_sink.ex b/lib/sequin/consumers/s2_sink.ex
index fdd560e8..2a4dd839 100644
--- a/lib/sequin/consumers/s2_sink.ex
+++ b/lib/sequin/consumers/s2_sink.ex
@@ -7,7 +7,7 @@ defmodule Sequin.Consumers.S2Sink do

   alias Sequin.Encrypted

-  @derive {Jason.Encoder, only: [:basin, :stream]}
+  @derive {Jason.Encoder, only: [:basin, :stream, :dashboard_url]}
   @derive {Inspect, except: [:access_token]}
   @primary_key false
   typed_embedded_schema do
@@ -15,17 +15,23 @@ defmodule Sequin.Consumers.S2Sink do
     field :basin, :string
     field :stream, :string
     field :access_token, Encrypted.Field
+    field :dashboard_url, :string, virtual: true
   end

   def endpoint_url(%__MODULE__{basin: basin}) do
     "https://#{basin}.b.aws.s2.dev/v1/"
   end

+  def dashboard_url(%__MODULE__{basin: basin}) do
+    "https://s2.dev/dashboard/basins/#{basin}"
+  end
+
   def changeset(struct, params) do
     struct
     |> cast(params, [:basin, :stream, :access_token])
     |> validate_required([:basin, :stream, :access_token])
     |> validate_basin()
+    |> put_change(:dashboard_url, dashboard_url(struct))
   end
```

acco: @davoclavo I agree adding it as a virtual attr is a little strange - but fortunately, I don't think we need to!

When we're serializing/deserializing JSON, we're fans of doing it explicitly as map literals. So, you can see here where we encode the SinkConsumer.sink struct for the frontend:

https://github.com/sequinstream/sequin/blob/e11a1cf5d7840d4fb5315da76033e8a46574be2d/lib/sequin_web/live/sink_consumers/show.ex#L826

We do this so we can abstract the frontend data structures from the backend data structures however we want

lib/sequin/sinks/s2/client.ex
```diff
@@ -0,0 +1,101 @@
+defmodule Sequin.Sinks.S2.Client do
+  @moduledoc false
+  alias Sequin.Consumers.S2Sink
+  alias Sequin.Error
+  alias Sequin.Sinks.S2.HttpClient
+
+  require Logger
+
+  @spec test_connection(S2Sink.t()) :: :ok | {:error, Error.t()}
+  def test_connection(%S2Sink{} = sink) do
```
acco, line 43: Great error messages

lib/sequin/sinks/s2/http_client.ex
```diff
@@ -0,0 +1,39 @@
+defmodule Sequin.Sinks.S2.HttpClient do
```
acco: All good stuff in here - you might consider just collapsing into Client

test/sequin/s2_client_test.exs
```diff
@@ -0,0 +1,36 @@
...
+  alias Sequin.Sinks.S2.Client
+  alias Sequin.Sinks.S2.HttpClient
+
+  @sink %S2Sink{type: :s2, basin: "test", stream: "test-stream", access_token: "token"}
+
+  describe "append_records/2" do
+    test "successfully sends batch" do
+      records = [SinkFactory.s2_record(), SinkFactory.s2_record()]
+
+      Req.Test.stub(HttpClient, fn conn ->
```
acco:
```suggestion
      Req.Test.expect(HttpClient, fn conn ->
```

acco (commented):
Looking good so far!

Addressing your questions @davoclavo :

> Should we handle propagating timestamps differently - especially when we are doing a backfill?
> Source: https://s2.dev/docs/rest/records/append-records#body-records-timestamp

Hmm... I'm not sure what we should do for timestamps! I'll ask their team now

> How to handle cases where the batch size exceeds the 1MiB that is required by S2?

Unfortunately, we won't know batch size until we make the JSON. And in Broadway (pipeline we're using to make batches) we can't say "give me a batch that has this many items OR is of this size, whichever is lesser". So we may get an extra large batch in `handle_batch`

The easiest/dumbest way to do it: if the batch is rejected for being too big, retry each message in the batch individually. We do this here (this pipeline is for a secret-menu feature of Sequin -- don't worry about context too much!):

https://github.com/sequinstream/sequin/blob/e11a1cf5d7840d4fb5315da76033e8a46574be2d/lib/sequin/runtime/http_push_pipeline.ex#L374-L388

This is ~fine for a batch of 10 messages, kinda falls apart at 100.

So, I'd say either:

1. Let's start with batch_size=1 to avoid this issue. Not the highest throughput, but OK to start.
2. Let's start with a small batch_size (like 10, not configurable) that minimizes chance of this issue and lets us do an easy fallback if we hit the limit

If an individual message is too big for S2, we can [silently drop it for now](https://github.com/sequinstream/sequin/issues/937) - or just fail it indefinitely.

lib/sequin/consumers/s2_sink.ex
```diff
@@ -0,0 +1,41 @@
...
+
+  defp validate_basin(changeset) do
+    changeset
+    |> validate_change(:basin, fn :basin, basin ->
+      case basin do
+        basin when is_binary(basin) and basin != "" -> []
+        _ -> [basin: "must be a non-empty string"]
+      end
+    end)
+    |> validate_length(:basin, max: 255)
```
shikhar: if you'd like to make it more precise:
- between 8 & 48 chars
- `r"^[a-z0-9]([a-z0-9-]*[a-z0-9])?$"`

lib/sequin/sinks/s2/client.ex
shikhar: agreed, thanks for adding useful diagnostics!

shikhar:
> According to their docs only gzip compression for request body is supported

zstd too!

test/sequin/s2_client_test.exs
davoclavo: Oh thanks for re-raising this suggestion!

Initially I didn't notice any change (e.g If I commented out the line where the request should happen, the test still passed), but after a bit of digging I noticed there is something odd and unexpected. It seems that the [call to `Req.Test.verify_on_exit!`](https://github.com/sequinstream/sequin/blob/codex/add-s2.dev-sink/test/support/case.ex#L51) and the call to [`Hammock.verify_on_exit!`](https://github.com/sequinstream/sequin/blob/codex/add-s2.dev-sink/test/support/case.ex#L23) are stepping on eachother toes, probably because both use `Mox` under the hood.

If I comment out the `Hammock.verify_on_exit!` call, the test indeed fails if when the expected request is not sent.

Probably this is out of scope for this PR, but we should fix that issue; as there are probably tests where the expected behavior of `Req.Test.expect` is not actually happening.

assets/svelte/sinks/s2/S2SinkCard.svelte
davoclavo: [Done](https://github.com/sequinstream/sequin/pull/1795/commits/39e7b67173cdb66c6da14ab4b8e5a30dd11e8352), thanks both for your feedback!

test/sequin/s2_client_test.exs
davoclavo: #1799

lib/sequin/runtime/s2_pipeline.ex
davoclavo: Just to be clear why I resolved this comment:

S2 expects the body to be a string, so we need to `Jason.encode!` here to make sure the object is sent as a string, and not as a nested JSON object

lib/sequin_web/live/components/consumer_form.ex
```diff
@@ -981,11 +1027,11 @@ defmodule SequinWeb.Components.ConsumerForm do
 
     %{
       "type" => "gcp_pubsub",
-      "project_id" => sink.project_id,
-      "topic_id" => sink.topic_id,
-      "use_emulator" => sink.use_emulator,
-      "emulator_base_url" => sink.emulator_base_url,
-      "credentials" => Jason.encode!(Sequin.Map.reject_nil_values(creds), pretty: true)
+      "project_id" => sink["project_id"],
```
davoclavo, line 1030: Remove these changes

PR #1803, opened 2025-06-13
:recycle: Move flushing out of handle_data

acco:
Move all flushing logic to one location. `handle_data` ought to focus on message accumulation. Inside of `:flush_messages` (timer), we can do the flushing. Now, we'll check after receiving messages if we should flush right away. If we should, we'll reschedule the timer for 0.

This is in preperation for soft disconnects/stopping receives from tcp.

### DIFF
--- a/lib/sequin/runtime/slot_processor_server.ex
+++ b/lib/sequin/runtime/slot_processor_server.ex
@@ -11,8 +11,6 @@ defmodule Sequin.Runtime.SlotProcessorServer do
 
   use Sequin.ProcessMetrics.Decorator
 
-  import Sequin.Error.Guards, only: [is_error: 1]
-
   alias __MODULE__
   alias Ecto.Adapters.SQL.Sandbox
   alias Sequin.Constants
@@ -156,7 +154,6 @@ defmodule Sequin.Runtime.SlotProcessorServer do
       field :message_store_refs, %{SinkConsumer.id() => reference()}, default: %{}
 
       # Health and monitoring
-      field :bytes_received_since_last_limit_check, non_neg_integer(), default: 0
       field :check_memory_fn, nil | (-> non_neg_integer())
       field :dirty, boolean(), default: false
 
@@ -422,7 +419,6 @@ defmodule Sequin.Runtime.SlotProcessorServer do
     raw_bytes_received = byte_size(msg)
     ProcessMetrics.increment_throughput("raw_bytes_received", raw_bytes_received)
 
-    state = maybe_schedule_flush(state)
     state = %{state | message_received_since_last_heartbeat: true}
 
     state =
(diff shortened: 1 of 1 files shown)


acco:
* **#1804** <a href="https://app.graphite.dev/github/pr/sequinstream/sequin/1804?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1803** <a href="https://app.graphite.dev/github/pr/sequinstream/sequin/1803?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/sequinstream/sequin/1803?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

lib/sequin/runtime/slot_processor_server.ex
```diff
@@ -156,7 +154,6 @@ defmodule Sequin.Runtime.SlotProcessorServer do
       field :message_store_refs, %{SinkConsumer.id() => reference()}, default: %{}
 
       # Health and monitoring
-      field :bytes_received_since_last_limit_check, non_neg_integer(), default: 0
```
acco, line 159: This was previously used to ensure we don't check limit on every message receive, only on an interval (the interval being some number of bytes). We don't need this anymore, as we only check the limit before flushing.

lib/sequin/runtime/slot_processor_server.ex
```diff
@@ -422,57 +419,19 @@ defmodule Sequin.Runtime.SlotProcessorServer do
...
-
-      error = if is_error(e), do: e, else: Error.service(service: :replication, message: Exception.message(e))
-
-      Health.put_event(
-        state.replication_slot,
-        %Event{slug: :replication_message_processed, status: :fail, error: error}
-      )
-
-      reraise e, __STACKTRACE__
+    {:keep_state, state}
```
acco, line 434: Don't flush in handle_data. Only in `handle_info(:flush_messages)`, which is called on a timer. `maybe_schedule_flush` is more sophisticated now and will schedule a flush right away if we've hit byte threshold

lib/sequin/runtime/slot_processor_server.ex
```diff
@@ -583,14 +542,33 @@ defmodule Sequin.Runtime.SlotProcessorServer do
...
+        )
+
+        {:ok, state} = flush_messages(state)
+
+        if state.test_pid do
+          send(state.test_pid, {:stop_replication, state.id})
+        end
+
+        {:disconnect, :over_system_memory_limit, state}
+
```
acco, line 571: This else clause moved from the `handle_data` flush site -- handles `check_limit`

I'm not sure the other clauses or rescue were in use!

lib/sequin/runtime/slot_processor_server.ex
```diff
@@ -1683,26 +1681,20 @@ defmodule Sequin.Runtime.SlotProcessorServer do
...
-        {:ok, %{state | bytes_received_since_last_limit_check: 0}}
-      end
+    if current_memory >= state.max_memory_bytes do
+      Process.sleep(@check_limit_interval)
+      check_limit(state, attempt + 1)
     else
-      new_bytes = state.bytes_received_since_last_limit_check + raw_bytes_received
-      {:ok, %{state | bytes_received_since_last_limit_check: new_bytes}}
+      :ok
     end
```
acco, line 1698: simplify check_limit, as it's called on an interval now

These sleeps/backoffs can probably be refactored out soon, as we will no longer be disconnecting when limits are hit

PR #1804, opened 2025-06-13
:recycle: Buffer messages as opposed to disconnecting

acco:
To back-pressure, we can simply not read messages in from the tcp socket. However, we don't own the socket - Postgrex.Protocol does. When we call `handle_copy_recv`, that will flip on `active: :once` on the socket after reading in all messages. So, our easiest strategy is to simply not call `handle_copy_recv` anymore if we're not ready for new messages.

This means to stop the flow, up in `ReplicationConnection`, we flip on buffering to buffer new tcp messages in state. Only when the pipeline indicates it's ready for more do we process the buffer and resume processing tcp messages as they come in.

### DIFF
--- a/lib/sequin/postgres/replication_connection.ex
+++ b/lib/sequin/postgres/replication_connection.ex
@@ -152,7 +152,9 @@ defmodule Sequin.Postgres.ReplicationConnection do
   defstruct protocol: nil,
             state: nil,
             reconnect_backoff: 500,
-            streaming: nil
+            streaming: nil,
+            buffering_sock_messages?: false,
+            buffered_sock_messages: []
 
   ## PUBLIC API ##
 
@@ -237,6 +239,8 @@ defmodule Sequin.Postgres.ReplicationConnection do
               | {:query, query, state}
               | {:stream, query, stream_opts, state}
               | {:disconnect, reason, state}
+              | {:inactivate_socket, state}
+              | {:activate_socket, state}
 
   @doc """
   Callback for `call/3`.
@@ -437,6 +441,9 @@ defmodule Sequin.Postgres.ReplicationConnection do
   @impl :gen_statem
   def callback_mode, do: :handle_event_function
 
+  defguardp is_socket_message(message)
+            when elem(message, 0) in [:tcp, :tcp_closed, :tcp_error, :ssl, :ssl_closed, :ssl_error]
+
   @doc false
   @impl :gen_statem
   def init({mod, arg, opts}) do
@@ -454,7 +461,8 @@ defmodule Sequin.Postgres.ReplicationConnection do
 
         state = %__MODULE__{
           reconnect_backoff: reconnect_backoff,
-          state: {mod, mod_state}
+          state: {mod, mod_state},
+          buffered_sock_messages: []
         }
 
         put_opts(opts)
@@ -493,18 +501,24 @@ defmodule Sequin.Postgres.ReplicationConnection do
     handle(mod, :handle_call, [msg, from, mod_state], from, s)
   end
 
-  def handle_event(:info, msg, @state, %{protocol: protocol, streaming: streaming} = s) do
+  def handle_event(:info, msg, @state, %__MODULE__{buffering_sock_messages?: true} = s) when is_socket_message(msg) do
+    {:keep_state, %{s | buffered_sock_messages: [msg | s.buffered_sock_messages]}}
+  end
+
+  def handle_event(:info, msg, @state, %__MODULE__{protocol: protocol, streaming: streaming} = s)
+      when is_socket_message(msg) do
     case Protocol.handle_copy_recv(msg, streaming, protocol) do
       {:ok, copies, protocol} ->
         handle_data(copies, %{s | protocol: protocol})
 
-      :unknown ->
-        %{state: {mod, mod_state}} = s
-        maybe_handle(mod, :handle_info, [msg, mod_state], s)
-
       {error, reason, protocol} ->
         reconnect_or_stop(error, reason, protocol, s)
     end
+  end
+
+  def handle_event(:info, msg, _state, s) do
+    %{state: {mod, mod_state}} = s
+    maybe_handle(mod, :handle_info, [msg, mod_state], s)
   end
 
   def handle_event({:timeout, :reconnect}, nil, @state, s) do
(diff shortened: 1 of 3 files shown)


acco:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/sequinstream/sequin/1804?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1804** <a href="https://app.graphite.dev/github/pr/sequinstream/sequin/1804?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/sequinstream/sequin/1804?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1803** <a href="https://app.graphite.dev/github/pr/sequinstream/sequin/1803?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

lib/sequin/postgres/replication_connection.ex
```diff
@@ -237,6 +239,8 @@ defmodule Sequin.Postgres.ReplicationConnection do
               | {:query, query, state}
               | {:stream, query, stream_opts, state}
               | {:disconnect, reason, state}
+              | {:inactivate_socket, state}
+              | {:activate_socket, state}
```
acco, line 243: We need to expand the API - this is for `handle_info`. 

Only one function in SlotProcessorServer will return these tuples, `handle_info(:flush_messages)`. It will return `:activate` all the time during normal operation. It will return `inactivate` if limits are hit.

lib/sequin/postgres/replication_connection.ex
```diff
@@ -437,6 +441,9 @@ defmodule Sequin.Postgres.ReplicationConnection do
   @impl :gen_statem
   def callback_mode, do: :handle_event_function
 
+  defguardp is_socket_message(message)
+            when elem(message, 0) in [:tcp, :tcp_closed, :tcp_error, :ssl, :ssl_closed, :ssl_error]
```
acco, line 445: If you'll recall, `Protocol.handle_copy_recv` would just return `:unknown` for non-protocol messages! So all handle_infos passed through Protocol first.

We need to know earlier on if it's a protocol/socket message, so we have this guard up here.

acco, line 445: If you'll recall, `Protocol.handle_copy_recv` would just return `:unknown` for non-protocol messages! So all handle_infos passed through Protocol first.

We need to know earlier on if it's a protocol/socket message, so we have this guard up here.

lib/sequin/postgres/replication_connection.ex
```diff
@@ -493,20 +501,26 @@ defmodule Sequin.Postgres.ReplicationConnection do
     handle(mod, :handle_call, [msg, from, mod_state], from, s)
   end
 
-  def handle_event(:info, msg, @state, %{protocol: protocol, streaming: streaming} = s) do
+  def handle_event(:info, msg, @state, %__MODULE__{buffering_sock_messages?: true} = s) when is_socket_message(msg) do
+    {:keep_state, %{s | buffered_sock_messages: [msg | s.buffered_sock_messages]}}
+  end
```
acco, line 506: We have to capture new incoming tcp messages and buffer them - we can't let them hit the next clause and call Protocol.handle_copy_recv, because that will re-activate the socket after producing all messages

lib/sequin/postgres/replication_connection.ex
```diff
@@ -603,7 +638,12 @@ defmodule Sequin.Postgres.ReplicationConnection do
     # Exception is unused
     Protocol.disconnect(%RuntimeError{}, protocol)
 
-    maybe_handle(mod, :handle_disconnect, [error, reason, mod_state], %{s | protocol: nil, streaming: nil})
+    maybe_handle(mod, :handle_disconnect, [error, reason, mod_state], %__MODULE__{
+      s
+      | protocol: nil,
+        streaming: nil,
+        buffered_sock_messages: []
```
acco, line 645: On disconnect, need to clear the buffer - pretty sure this is the only spot!

lib/sequin/runtime/slot_processor_server.ex
```diff
@@ -82,6 +83,20 @@ defmodule Sequin.Runtime.SlotProcessorServer do
...
+        if Application.get_env(:sequin, :env) == :test do
+          5
+        else
+          :timer.seconds(1)
+        end
+
+      value ->
+        value
+    end
+  end
```
acco, line 98: From my experimenting, this should be a little longer, to let the system drain

lib/sequin/postgres/replication_connection.ex
```diff
@@ -493,20 +501,26 @@ defmodule Sequin.Postgres.ReplicationConnection do
...
-
       {error, reason, protocol} ->
         reconnect_or_stop(error, reason, protocol, s)
     end
   end
 
+  def handle_event(:info, msg, _state, s) do
+    %{state: {mod, mod_state}} = s
+    maybe_handle(mod, :handle_info, [msg, mod_state], s)
+  end
```
acco, line 522: This is for "regular' handle info calls

We can't buffer *any* handle info call, as that would obviously buffer all of our timers and other things. So those bypass the is_socket_message guard clause and do their thing down here

lib/sequin/postgres/replication_connection.ex
```diff
@@ -553,6 +567,14 @@ defmodule Sequin.Postgres.ReplicationConnection do
           {error, reason, protocol} -> reconnect_or_stop(error, reason, protocol, s)
         end
 
+      {:inactivate_socket, mod_state} ->
+        s = %__MODULE__{s | buffering_sock_messages?: true, state: {mod, mod_state}}
+        {:keep_state, s}
+
+      {:activate_socket, mod_state} ->
+        s = maybe_flush_buffered_messages(s)
+        {:keep_state, %{s | state: {mod, mod_state}}}
```
acco, line 576: new possible tuple returns

lib/sequin/postgres/replication_connection.ex
acco, line 506: We have to capture new incoming tcp messages and buffer them - we can't let them hit the next clause and call Protocol.handle_copy_recv, because that will re-activate the socket after producing all messages

lib/sequin/postgres/replication_connection.ex
```diff
@@ -591,6 +613,19 @@ defmodule Sequin.Postgres.ReplicationConnection do
...
+  defp maybe_flush_buffered_messages(%__MODULE__{buffering_sock_messages?: true, buffered_sock_messages: []} = s),
+    do: %{s | buffering_sock_messages?: false}
+
+  defp maybe_flush_buffered_messages(%__MODULE__{buffering_sock_messages?: true} = s) do
+    s.buffered_sock_messages
+    |> Enum.reverse()
+    |> Enum.each(fn msg -> send(self(), msg) end)
+
+    %{s | buffering_sock_messages?: false, buffered_sock_messages: []}
+  end
```
RTLS, line 625: is this ordering safe with messages from the TCP socket in our mailbox? active once?

PR #1815, opened 2025-06-14
Escape table schema and name in replica identity full query

acco:


RTLS:
I already merged to main

acco:
Thanks @RTLS !

Issue #1371, opened 2025-05-06
Add Meilisearch as a native destination


thisisgoldman:
## Description

Request to add Meilisearch (https://github.com/meilisearch/meilisearch) as a native destination in Sequin. This would allow users to stream database changes directly to Meilisearch search indexes, enabling automatic synchronization of search indexes with Postgres data.

## Use Case

This integration would be valuable for:
- Keeping search indexes automatically in sync with database changes
- Eliminating the need for custom sync code between Postgres and Meilisearch
- Supporting real-time search experiences by leveraging Sequin's low-latency CDC capabilities
- Allowing users to benefit from Meilisearch's features (typo tolerance, faceted search, hybrid search) with Postgres data

rhodee:
@thisisgoldman I achieved reactive delivery with a transform, dynamic routing, using a webhook destination. 

Here is my brief experience report:

I was able to get it all working, making a few discoveries. After upgrading to `v0.8.0` I could not get Meilisearch to accept my payloads. The logs routinely mentioned it had `0 successful tasks and 1 failing tasks` after every message Sequin delivered. On my webhook sink, I toggled off batch mode. I did this because after reading the Meilisearch GET `/tasks` route, it was complaining it could not find my primaryKey due to the body I was sending. This is the object it received:

```ts
{
  data: []
}
```

@acco it may be less surprising to have an array of objects instead of an array of objects hanging off the `data` field, but I leave that decision to your team. If it were configurable, that could be nice. I had assumed the payload I returned from the function _is_ what is sent, and in most cases, I assume it is. 

Reference: https://github.com/sequinstream/sequin/issues/1194

Good luck @thisisgoldman!

acco:
Hey @rhodee – thanks so much for sharing this report! It's so useful to hear this stuff

Totally understood re: the batch confusion. We only recently introduced the "batch" toggle for webhooks, to remove the outer `data` envelope. If we were starting today, we'd consider inverting the default (don't batch by default). But we want to minimize breaking changes, so you have to explicitly turn batch mode off.

Now with transforms, routing, and batch_mode=false, you should have a lot of control around crafting your webhook sink requests

bugproof:
Are there any docs explaining how to write your own sinks? I'm interested in native meilisearch destination as well. 

Are websockets good for this in the meantime? What if I miss some changes e.g. due to some network outage/crash? How do I backfill?

acco:
[...]

Issue #1194, opened 2025-04-03
Transform HTTPEndpoint


rhodee:
I am trying to integrate [Svix](https://docs.svix.com/quickstart#send-a-message) as a webhook sink within Sequin.

The document I linked to shows how to send a webhook via API. Wildcard support is needed.

The cURL payload looks like this:

```sh
curl -X POST "https://api.svix.com/api/v1/app/app_Xzx8bQeOB1D1XEYmAJaRGoj0/msg/" \
    -H  "Accept: application/json" \
    -H  "Content-Type: application/json" \
    -H  "Authorization: Bearer AUTH_TOKEN" \
    -d '{"eventType": "invoice.paid", "eventId": "evt_Wqb1k73rXprtTm7Qdlr38G", "payload": {"type": "event.type", "id": "invoice_WF7WtCLFFtd8ubcTgboSFNql", "status": "paid", "attempt": 2}}'
```

Would this be possible with the transforms API?

acco:
@rhodee Just confirming - do you just need to transform the body? Or does your path need to be dynamic at all based on the message?

rhodee:
Hey @acco - the body must remain unchanged. But the route to send the webhook, path must change. I need to be able to specify something akin to this image. I'd like to use the entity ID in the path for the webhook. Nothing else is needed for this to be 🧑‍🍳 

<img width="817" alt="Image" src="https://github.com/user-attachments/assets/6ad96761-10e6-4f0a-9df6-3ef66ff13359" />

acco:
Thanks @rhodee!

To summarize, you're bringing up two things:

1. **The ability to modify the request path**: This is a very cool idea and one that we're literally whiteboarding in a conference room right now :) This is the next thing we're going to work on
2. **Adding query params**: Our webhook sink does not support hard-coding query params at all

Is that right?

For #2 -- would it be worthwhile for us to add that support ahead of adding #1? #1 is going to take us a little time (we want to ship this month), but #2 we can add relatively quickly.

rhodee:
@acco, thanks for this feedback. I'm most interested in #1. It unlocks a direct integration with Svix, a tool many companies are using.

The query params feature is not a showstopper. Specifically, Meilisearch supports the use of a query param, or it will infer from the payload. What _would_ make Sequin to Meilisearch work seamlessly is being able to add documents as a webhook sink.

[...]

PR #1823, opened 2025-06-16
add Meilisearch sink

fcannizzaro:
Hi there! 👋
This PR adds a new Meilisearch sink, based on the existing Typesense sink. It includes both the implementation and the docs.

It’s actually my first time writing Elixir, so if anything looks a bit off, feel free to point it out 😄.

**Related Issues**
#1371, #1194

Thanks, and looking forward to your thoughts!

### DIFF
--- a/lib/sequin/sinks/meilisearch/client.ex
+++ b/lib/sequin/sinks/meilisearch/client.ex
@@ -23,7 +23,7 @@ defmodule Sequin.Sinks.Meilisearch.Client do
         {:ok, %{body: body}} ->
           case body do
             %{"status" => status} when status in ["enqueued", "processing"] ->
-              timeout = Sequin.Time.exponential_backoff(50, retries, 10_000)
+              timeout = Sequin.Time.exponential_backoff(200, retries, 10_000)
               Logger.warning("[Meilisearch] Task #{task_id} is still in progress (#{retries}/10)")
               :timer.sleep(timeout)
               verify_task_by_id(sink, task_id, retries + 1)
--- a/lib/sequin/transforms/transforms.ex
+++ b/lib/sequin/transforms/transforms.ex
@@ -1187,19 +1187,19 @@ defmodule Sequin.Transforms do
      }}
   end
 
-    # Add parse_sink for meilisearch type
-    defp parse_sink(%{"type" => "meilisearch"} = attrs, _resources) do
-      {:ok,
-       %{
-         type: :meilisearch,
-         endpoint_url: attrs["endpoint_url"],
-         index_name: attrs["index_name"],
-         primary_key: attrs["primary_key"],
-         api_key: attrs["api_key"],
-         batch_size: attrs["batch_size"],
-         timeout_seconds: attrs["timeout_seconds"]
-       }}
-    end
+  # Add parse_sink for meilisearch type
+  defp parse_sink(%{"type" => "meilisearch"} = attrs, _resources) do
+    {:ok,
+     %{
+       type: :meilisearch,
+       endpoint_url: attrs["endpoint_url"],
+       index_name: attrs["index_name"],
+       primary_key: attrs["primary_key"],
+       api_key: attrs["api_key"],
+       batch_size: attrs["batch_size"],
+       timeout_seconds: attrs["timeout_seconds"]
+     }}
+  end
 
   # Add parse_sink for elasticsearch type
   defp parse_sink(%{"type" => "elasticsearch"} = attrs, _resources) do
--- a/test/support/factory/character_factory.ex
+++ b/test/support/factory/character_factory.ex
@@ -35,7 +35,7 @@ defmodule Sequin.Factory.CharacterFactory do
     attrs
     |> character()
     |> Sequin.Map.from_ecto()
-    |> Enum.into(%{}, fn {k, v} -> {to_string(k), v} end)
+    |> Map.new(fn {k, v} -> {to_string(k), v} end)
   end
 
   def insert_character!(attrs \\ [], opts \\ []) do


acco:
Wow, thanks for opening this @fcannizzaro ! We'll take a look soon

fcannizzaro:
Yeah usually I use the 'id' field so I forgot about it. I think we can add it as param in the configuration of the sink.

docs/how-to/stream-postgres-to-meilisearch.mdx
```diff
@@ -0,0 +1,151 @@
...
+
+  <Step title="Transforms">
+    Meilisearch requires a defined schema with:
+    1. A unique document `id` (string or integer).
+    2. Searchable and filterable fields.
+
+    Use a [functional transform](/reference/transforms#function-transform) to convert Postgres rows to Meilisearch documents:
+
+    ```elixir
+    def transform(record) do
```
davoclavo: Update all the example transform function signatures to include all the other required arguments: `action, record, changes, metadata`


```elixir
def transform(action, record, changes, metadata) do
  ...
end
```

lib/sequin/sinks/meilisearch/client.ex
```diff
@@ -0,0 +1,267 @@
...
+
+    case Req.get(req, url: "/indexes/#{index_name}") do
+      {:ok, %{status: st, body: body}} when st in 200..299 ->
+        {:ok, body}
+
+      {:ok, %{status: 404}} ->
+        {:error,
+         Error.service(
+           service: :meilisearch,
+           message: "Index '#{index_name}' not found"
```
davoclavo: I think it's fine to start with this strict implementation requiring the index to already exist. 

However meilisearch supports implicitly creating indexes on the first document insertion.

https://www.meilisearch.com/docs/learn/getting_started/indexes#implicit-index-creation

lib/sequin/runtime/meilisearch_pipeline.ex
```diff
@@ -0,0 +1,103 @@
...
+  def handle_batch(:default, messages, _batch_info, context) do
+    %{
+      consumer: %SinkConsumer{sink: sink} = consumer,
+      meilisearch_client: client,
+      test_pid: test_pid
+    } = context
+
+    setup_allowances(test_pid)
+
+    records = Enum.map(messages, fn %{data: data} -> data.data.record end)
```
davoclavo: We should add a call to `Message.to_external` to actually run the configured transform function, otherwise messages never get transformed.

```diff
+alias Sequin.Transforms.Message
...    
-        records = Enum.map(messages, fn %{data: data} -> data.data.record end)
+        records = Enum.map(messages, fn %{data: message} ->
+           Message.to_external(consumer, message)
+        end)
```

same comment applies for the `handle_batch(:delete, ...)`

lib/sequin/sinks/meilisearch/client.ex
```diff
@@ -0,0 +1,267 @@
...
+
+  @doc """
+  Import multiple documents in JSONL format.
+  """
+  def import_documents(%SinkConsumer{} = consumer, %__MODULE__{} = client, index_name, jsonl) do
+    req =
+      client
+      |> base_request()
+      |> Req.merge(
+        url: "/indexes/#{index_name}/documents",
```
davoclavo: We need to confirm for the batch to be confirmed until we can successfully return from this function.

If documents fail inserting, we should report that error so we properly report that a message was not sent successfully.

For example, with this implementation if documents are missing the `primaryKey`, the batch insert request returns a 202, but the actual insertion that happens asynchronously a bit later may fail. And we are erroneously reporting that the insertion succeeded.

So after inserting the batch, we should parse the response, get the `taskUid` and poll until the batch is complete and report the results. 

https://www.meilisearch.com/docs/reference/api/tasks#get-one-task


Question for @acco @RTLS: is there a mechanism for these async batch confirmations within Sequin pipelines? or perhaps an existing module that easily handles waiting/polling for an async job?

---

Another option is to use a synchronous-insert endpoint to avoid having to deal with this complexity for now, but I couldn't find anything within the mailisearch docs that show it supports that

(This comment also applies for the delete functionality I think)

lib/sequin/sinks/meilisearch/client.ex
```diff
@@ -0,0 +1,267 @@
...
+    end
+  end
+
+  # Private helpers
+
+  defp base_request(%__MODULE__{} = client) do
+    Req.new(
+      base_url: client.url,
+      headers: [{"Authorization", "Bearer #{client.api_key}"}],
+      receive_timeout: :timer.seconds(client.timeout_seconds),
```
davoclavo: add `compress_body: true` setting so we compress the payload with gzip

docs/reference/sinks/meilisearch.mdx
```diff
@@ -0,0 +1,74 @@
...
+
+- **API key**
+
+    The API key for authenticating with your Meilisearch server.
+
+## Transform requirements
+
+Your [transform](/reference/transforms) must return a document matching the schema of the [Meilisearch index](https://www.meilisearch.com/docs/learn/core_concepts/documents).
+
+This includes a primary key field (usually called `id`) which is mandatory for Meilisearch.
```
davoclavo: Maybe adjust the docs to state that the primary key should be at the *top level* of the map returned by the transform function, or does meilisearch support definining the primary key as a path? e.g. (`record.id`)

lib/sequin/sinks/meilisearch/client.ex
fcannizzaro: For this problem I was writing the task verification right now, after insertion/deletion I'm checking the status of the task until is succeeded or failed (`L25` of client.ex)

lib/sequin/runtime/meilisearch_pipeline.ex
fcannizzaro: Ah, sorry, I didn't understand the use of Message.to_external(). I'll fix that now.

docs/reference/sinks/meilisearch.mdx
fcannizzaro: I changed that paragraph to:

For example, the following transform ensures the primary key (`id`) is at the top level of the returned map:

```elixir
def transform(action, record, changes, metadata) do
  Map.put(record, "id", Kernel.to_string(record["product"]["id"]))
end
```

lib/sequin/sinks/meilisearch/client.ex
fcannizzaro: Yeah, I removed that error, since the index is created if not exist. The `get_index` is still used to "Test connection".

fcannizzaro:
Ignore this latest force-push commit, I accidentally rebased main into my branch and rolled back... anyway on local now `mix test` passes.

lib/sequin/sinks/meilisearch/client.ex
```diff
@@ -0,0 +1,244 @@
...
+    end
+  end
+
+  @doc """
+  Get information about an index.
+  """
+  def get_index(%__MODULE__{} = client, index_name) do
+    req = base_request(client)
+
+    case Req.get(req, url: "/indexes/#{index_name}") do
```
davoclavo: If a user provides a stream name that does not exist the server is currently crashing with this error:


```
[error] [12:16:28.944] GenServer #PID<0.4070.0> terminating
** (CaseClauseError) no case clause matching: {:ok, %Req.Response{status: 404, headers: %{"content-type" => ["application/json"], "date" => ["Wed, 18 Jun 2025 18:16:28 GMT"], "transfer-encoding" => ["chunked"], "vary" => ["Origin, Access-Control-Request-Method, Access-Control-Request-Headers", "accept-encoding"]}, body: %{"code" => "index_not_found", "link" => "https://docs.meilisearch.com/errors#index_not_found", "message" => "Index `test-sequin-twoo` not found.", "type" => "invalid_request"}, trailers: %{}, private: %{}}}
    (sequin 0.1.0) lib/sequin/sinks/meilisearch/client.ex:185: Sequin.Sinks.Meilisearch.Client.get_index/2
    (sequin 0.1.0) lib/sequin_web/live/components/consumer_form.ex:564: SequinWeb.Components.ConsumerForm.test_meilisearch_connection/1
    (sequin 0.1.0) lib/sequin_web/live/components/consumer_form.ex:292: SequinWeb.Components.ConsumerForm.handle_event/3
    (phoenix_live_view 0.20.17) lib/phoenix_live_view/channel.ex:735: anonymous fn/4 in Phoenix.LiveView.Channel.inner_component_handle_event/4
    (telemetry 1.3.0) /Users/david/Desktop/dev/sequin/deps/telemetry/src/telemetry.erl:324: :telemetry.span/3
    (phoenix_live_view 0.20.17) lib/phoenix_live_view/diff.ex:209: Phoenix.LiveView.Diff.write_component/4
    (phoenix_live_view 0.20.17) lib/phoenix_live_view/channel.ex:662: Phoenix.LiveView.Channel.component_handle/4
    (stdlib 6.2) gen_server.erl:2345: :gen_server.try_handle_info/3
    (stdlib 6.2) gen_server.erl:2433: :gen_server.handle_msg/6
    (stdlib 6.2) proc_lib.erl:329: :proc_lib.init_p_do_apply/3
Last message: %Phoenix.Socket.Message{topic: "lv:phx-GEo1fdwRuXO5ah4B", event: "event", payload: %{"cid" => 1, "event" => "test_connection", "type" => "hook", "value" => %{}}, ref: "106", join_ref: "47"} line=2646 pid=<0.4070.0> mfa=:gen_server.error_info/8
```

lib/sequin/runtime/meilisearch_pipeline.ex
```diff
@@ -0,0 +1,99 @@
...
+  def handle_batch(:default, messages, _batch_info, context) do
+    %{
+      consumer: %SinkConsumer{sink: sink} = consumer,
+      meilisearch_client: client,
+      test_pid: test_pid
+    } = context
+
+    setup_allowances(test_pid)
+
+    records = Enum.map(messages, &Message.to_external(consumer, &1.data))
```
davoclavo: Regarding your question about the default output of `Message.to_transform`:

By default, we push out that [default change message shape](https://sequinstream.com/docs/reference/messages#change-messages) for sinks, but users are able to modify that by providing a transform function.


Now, that default gets a bit tricky to work with meilisearch, as the object we send will contain nested objects by default, thus the primary key will very likely *not* be at the top level.

I just did some tests and meilisearch is able to accept primary keys specifying a path, so for instance `record.id` would be a valid primary key that would work with default messages.


So I am thinking of an approach that covers the four cases:

- The user **did not** provide a transform fn, and their primary key **is not** nested
  - We can attempt to search that primary key within the `record` struct, and if found, we push that key to the top level
- The user **did not** provide a transform fn, and their primary key **is** nested
  - We leave the message untouched, as the user probably provided a nested key on purpose
- The user **did** provide a transform fn, and their primary key **is not** nested
  - We leave the transformed message untouched
- The user **did** provide a transform fn, and their primary key is nested
  - We leave the transformed message untouched


```diff
--- a/lib/sequin/runtime/meilisearch_pipeline.ex
+++ b/lib/sequin/runtime/meilisearch_pipeline.ex
@@ -47,14 +47,19 @@ defmodule Sequin.Runtime.MeilisearchPipeline do
   @impl SinkPipeline
   def handle_batch(:default, messages, _batch_info, context) do
     %{
-      consumer: %SinkConsumer{sink: sink} = consumer,
+      consumer: %SinkConsumer{sink: sink, transform: transform} = consumer,
       meilisearch_client: client,
       test_pid: test_pid
     } = context

     setup_allowances(test_pid)

-    records = Enum.map(messages, &Message.to_external(consumer, &1.data))
+    records =
+      Enum.map(messages, fn %{data: message} ->
+        Message.to_external(consumer, message)
+        |> attempt_to_fill_primary_key(transform, sink.primary_key)
+      end)

     jsonl = encode_as_jsonl(records)

@@ -96,4 +101,21 @@ defmodule Sequin.Runtime.MeilisearchPipeline do
   defp setup_allowances(test_pid) do
     Mox.allow(Sequin.TestSupport.DateTimeMock, test_pid, self())
   end
+
+  # If the consumer does not have a transform, attempt to fill in missing primary key
+  # by extracting it from record, but only for simple (non-dotted) primary keys
+  defp attempt_to_fill_primary_key(message, nil, primary_key) do
+    if String.contains?(primary_key, ".") do
+      # If primary key contains a dot, it's a nested primary key - leave message untouched
+      message
+    else
+      # For simple primary keys, attempt to extract from record if not already present
+      case get_in(message, [:record, primary_key]) do
+        nil -> message
+        value -> Map.put(message, primary_key, value)
+      end
+    end
+  end
+
+  defp attempt_to_fill_primary_key(message, _transform, _primary_key), do: message
 end
```

lib/sequin/sinks/meilisearch/client.ex
```diff
@@ -0,0 +1,244 @@
...
+      {:ok, %{status: status}} when status in 200..299 ->
```
davoclavo: Btw, it would be nice if we checked if the existing sink contains the configured primary key

lib/sequin/sinks/meilisearch/client.ex
```diff
@@ -0,0 +1,244 @@
...
+      url: opts |> Keyword.fetch!(:url) |> String.trim_trailing("/"),
+      api_key: Keyword.fetch!(opts, :api_key),
+      timeout_seconds: Keyword.fetch!(opts, :timeout_seconds)
+    }
+  end
+
+  @doc """
+  Verify the status of a task by its ID.
+  """
+  def verify_task_by_id(%__MODULE__{} = client, task_id) do
```
davoclavo: Great work with this task monitoring mechanism!! Its working nice, reporting errors right away.

Regarding the retries, I think we should probably at least set a maximum timeout (or number of retries), and perhaps (not required) some backoff mechanism? otherwise this potentially will wait forever

If I recall correctly, we also have a fallback mechanism in case a batch fails we are able to retry messages individually, in case a single message within the batch is poisoned/malformed

cc @acco @RTLS -- do we have any prior art on handling this case, in order to be somewhat consistent with other pipelines?? or do you have any additional comments?

README.md
```diff
@@ -64,6 +64,7 @@ In our production benchmarks with AWS RDS Postgres and MSK Kafka, Sequin maintai
 | GCP Pub/Sub | ✅ Real-time streaming<br />✅ Backfill existing rows | Publish messages to Google Cloud Pub/Sub topics |
 | HTTP Pull (Native) | ✅ Real-time streaming<br />✅ Backfill existing rows | Consume changes directly from Sequin with exactly-once processing |
 | Kafka | ✅ Real-time streaming<br />✅ Backfill existing rows | Stream changes to Apache Kafka topics |
+| Meilisearch | ✅ Real-time streaming<br />✅ Backfill existing rows | Index database changes with Meilisearch |
```
RTLS, line 69: thank you for all the docs updates 🙏

lib/sequin/sinks/meilisearch/client.ex
RTLS: If async import is the only available endpoint then we must verify the task succeeds before we consider this import to be successful.

We don't currently have a generic way to await an async write

lib/sequin/sinks/meilisearch/client.ex
```diff
@@ -0,0 +1,244 @@
...
+    req = base_request(client)
+
+    case Req.get(req, url: "/tasks/#{task_id}") do
+      {:ok, %{body: body}} ->
+        case body do
+          %{"status" => status} when status in ["enqueued", "processing"] ->
+            :timer.sleep(1000)
+            verify_task_by_id(client, task_id)
```
RTLS: [Optional] One advanced improvement you could make is to check how long jobs are taking to process and report a health warning if they are taking excessively long. this is definitely not required, perhaps it's better to see if this is ever an issue before we build that.

Perhaps an easier place to start is to `Logger.warning` if they are taking a while -- but even that is not required to merge this PR imo

lib/sequin/sinks/meilisearch/client.ex
fcannizzaro: Yeah I forgot to limit the number of retries, anyway LTMK if there is something already implemented for this kind of needs

RTLS: there is no built in functionality for this at the moment. I think we should build a solution for meillisearch now and if/when we have multiple sinks that need this we can consider a generic behaviour.

your approach now seems good: perhaps just add exponential backoff and a limit. I don't have a good concept of how long meillisearch updates should take, but the maxiumn retry time should probably be quite generous

lib/sequin/sinks/meilisearch/client.ex
fcannizzaro: I made a batch insertion on postgres of `1M` of rows (2 columns only) with a simple transform. (executed on a M1 Pro)
Batch size: `10k`

Figure_1

From these results I think we should not never have any kind of issue anyway this should be good right ?

```diff
def verify_task_by_id(%__MODULE__{} = client, task_id, retries, timeout) do
+    if retries <= 0 do
+      {:error,
+     Error.service(
+         service: :meilisearch,
+         message: "Task verification timed out",
+         details: %{task_id: task_id}
+       )}
+    end

    req = base_request(client)

    case Req.get(req, url: "/tasks/#{task_id}") do
      {:ok, %{body: body}} ->
        case body do
          %{"status" => status} when status in ["enqueued", "processing"] ->
            Logger.warning(
              "[Meilisearch] Task #{task_id} is still in progress (status: #{status}). Retrying..."
            )
+            :timer.sleep(timeout)
+            verify_task_by_id(client, task_id, retries - 1, timeout * 2)

          %{"status" => "failed"} ->
            message = extract_error_message(body["error"])

            {:error,
             Error.service(
               service: :meilisearch,
               message: message,
               details: body
             )}

          _ ->
            {:ok}
        end

      {:error, reason} ->
        {:error, Error.service(service: :meilisearch, message: "Unknown error", details: reason)}
    end
  end

+  def verify_task_by_id(%__MODULE__{} = client, task_id) do
+    verify_task_by_id(client, task_id, 5, 1000)
+  end
```

RTLS: Thank you for testing this 🙏 based on your results I believe we should begin the exponential backoff at ~30 ms and have a max at ~1000ms. The reason to start with a fast backoff is to maximize throughput

Please use:

```elixir
defmodule Sequin.Time do
  @doc """
  Calculates exponential backoff time in milliseconds.

  ## Parameters
    - base: The base retry time in milliseconds (default: 1000)
    - count: The number of failed attempts (default: 0)
    - max: The maximum backoff time in milliseconds (default: 5 minutes)

  ## Examples
      iex> Sequin.Time.exponential_backoff(1000, 0)
      1000
      iex> Sequin.Time.exponential_backoff(1000, 3)
      8000
      iex> Sequin.Time.exponential_backoff(1000, 10, 60_000)
      60000
  """
  @spec exponential_backoff(non_neg_integer(), non_neg_integer(), non_neg_integer()) :: non_neg_integer()
  def exponential_backoff(base \\ :timer.seconds(1), count \\ 0, max \\ :timer.minutes(3)) do
    max_count = trunc(:math.log2(max / base) + 1)

    if count >= max_count do
      max
    else
      backoff = if count <= 1, do: base, else: base * :math.pow(2, count - 1)
      # Random value between 0.85 and 1.15
      jitter = :rand.uniform() * 0.3 + 0.85
      trunc(backoff * jitter)
    end
  end
end
```

ie.

```elixir
Sequin.Time.exponential_backoff(50, retry, 1000)
```

with perhaps 10 retries

```elixir
Enum.sum_by(1..5, &Sequin.Time.exponential_backoff(50, &1, 1000))
1751 # 1.7 seconds with 5 retries

Enum.sum_by(1..10, &Sequin.Time.exponential_backoff(50, &1, 1000))
6741 # 6.7 seconds with 10 retries
```

RTLS:
> * I executed all the format both for lib and assets dirs
> * `mix test` gives `4 doctests, 1053 tests, 1 failure, 13 skipped` but fails on some database files (?)
> 
> ```
>  1) test PostgresReplicationSlot end-to-end batch updates are delivered in sequence order via HTTP (Sequin.PostgresReplicationTest)
>      test/sequin/postgres_replication_test.exs:759
>      Did not receive :simple_http_server_loaded within 5 seconds
>      code: assert wait_and_validate_data(inspect(ref), -1, transactions_count * transaction_queries_count)
>      stacktrace:
>        test/sequin/postgres_replication_test.exs:789: (test)
> ```
> 
> I don't think it's related to this PR right ?

@fcannizzaro you can run `mix test --exclude jepsen` if this test is not passing for you -- it should not be related to your PR

lib/sequin/sinks/meilisearch/client.ex
davoclavo: Nice!

In order to reduce latency I perhaps would recommend to drastically reduce the sleep step to 25ms.

You could also use this function to calculate a similar backoff, but with a random jitter and a maximum timeout. (If you use this just make sure to increment your retries, instead of reducing them)

https://github.com/sequinstream/sequin/blob/main/lib/sequin/time.ex#L84-L112

```elixir
retries = 1
timeout = Sequin.Time.exponential_backoff(25, retries, 10_000)
# 25-ish ms
:timer.sleep(timeout)
...
retries = retries + 1
timeout = Sequin.Time.exponential_backoff(25, retries, 10_000)
# 50-ish ms
:timer.sleep(timeout)
...
```

EDIT:

Woops duplicated suggestion about `Sequin.Time.exponential_backoff`

NotAmaan:
This is awesome! Thank you @fcannizzaro and the reviewers for the hard work.
Meilisearch users don’t really have a decent alternative right now apart from building custom ETL pipelines. Really looking forward to trying this out as soon as it’s merged.

test/sequin/meilisearch_pipeline_test.exs
```diff
@@ -0,0 +1,182 @@
...
+    end
+
+    test "events are sent (no transform)", %{consumer: consumer} do
+      {ids, events} = generate_events(consumer, 2)
+
+      verify_network_request(consumer, events, fn message, i ->
+        # this works
+        message["record"]["id"] == Enum.at(ids, i)
+        # this does not work, because message["id"] is nil
+        # message["id"] == Enum.at(ids, i)
```
fcannizzaro: @davoclavo I have an issue with this test, caused by this line: 

```elixir
get_in(message, [:record, primary_key])
```

of `attempt_to_fill_primary_key` function.

This works fine at runtime, but is `nil` in the test environment. I think it's the type of structure that doest not support that kind of access.

 How can I fix it ?

davoclavo: I'll take a look at this soon!

davoclavo: @fcannizzaro after digging for a bit I think I found the issue:

Apparently in our Character factory (which generates entities from a test table), we are creating `record` maps and setting all the keys that get passed as `:atoms` instead of `"strings"` - thus when we do `get_in(message, [:record, primary_key])` no matches occur because `:id` is different than `"id"`. This behavior should not be present in production as it is not good practice to to dynamically generate atoms

So I suggest we have two options:

1. Option 1: Fixing out factories to use record keys as strings instead of atoms:

```diff
--- a/test/support/factory/character_factory.ex
+++ b/test/support/factory/character_factory.ex
@@ -35,6 +35,7 @@ defmodule Sequin.Factory.CharacterFactory do
     attrs
     |> character()
     |> Sequin.Map.from_ecto()
+    |> Enum.into(%{}, fn {k, v} -> {to_string(k), v} end)
   end
```

By the way, you will also have to adjust your tests to use string keys instead of atom keys

For instance:
```diff
-    ids = Enum.map(records, &Map.get(&1, :id))
+    ids = Enum.map(records, &Map.get(&1, "id"))
```
----

2. Option 2: Handling this edge case in the meilisearch pipeline implementation by attempting string keys first, and then atom keys (inside a try/catch via `String.to_existing_atom`)

```diff
--- a/lib/sequin/runtime/meilisearch_pipeline.ex
+++ b/lib/sequin/runtime/meilisearch_pipeline.ex
@@ -2,6 +2,8 @@ defmodule Sequin.Runtime.MeilisearchPipeline do
   @moduledoc false
   @behaviour Sequin.Runtime.SinkPipeline

+  require Logger
+
   alias Sequin.Consumers.SinkConsumer
   alias Sequin.Runtime.SinkPipeline
   alias Sequin.Runtime.Trace
@@ -119,7 +121,19 @@ defmodule Sequin.Runtime.MeilisearchPipeline do
     else
       # For simple primary keys, attempt to extract from record if not already present
       case get_in(message, [:record, primary_key]) do
-        nil -> message
+        nil ->
+          # Fallback check in case the record key is an atom (only should happen in tests)
+          try do
+            case get_in(message, [:record, String.to_existing_atom(primary_key)]) do
+              nil -> message
+              value -> Map.put(message, primary_key, value)
+            end
+          rescue
+            # In case atom does not already exist we fallback to the original message
+            ArgumentError ->
+                Logger.warn("Primary key '#{primary_key}' does not exist as an atom")
+              message
+          end
         value -> Map.put(message, primary_key, value)
       end
     end
```

---

**I think going for Option 1 for now is the best fix, as it would solve the issue at the root level**. Although if making that change breaks other unrelated tests perhaps going for Option 2 can be a fallback, and we could tackle fixing the factories as a separate issue

fcannizzaro: I followed the option 1, `mix test` is ok so no other test broke after the change

RTLS (approved):
Very nice work @fcannizzaro ! I just pulled it and everything looks and feels great. Thank you for your hard work opening this PR, taking into account all the great feedback from @davoclavo and others, and getting this new sink added 🙌

PR #1824, opened 2025-06-16
feat: support e2e sqs using localstack mock

abc3:


lib/sequin/consumers/sqs_sink.ex
```diff
@@ -18,15 +18,18 @@ defmodule Sequin.Consumers.SqsSink do
     field :access_key_id, :string
     field :secret_access_key, Encrypted.Field
     field :is_fifo, :boolean, default: false
+    field :endpoint, :string
+    field :proto, :string
```
acco: We use `scheme` elsewhere to match `URI`

lib/sequin/consumers/sqs_sink.ex
```diff
@@ -62,9 +65,18 @@ defmodule Sequin.Consumers.SqsSink do
...
+    client = AWS.Client.create(sink.access_key_id, sink.secret_access_key, sink.region)
+
+    client =
+      if sink.endpoint do
+        client
+        |> Map.put(:endpoint, sink.endpoint)
+        |> Map.put(:proto, sink.proto)
+      else
+        client
+      end
```
acco: 👍

priv/tests_e2e/docker-compose.yml
```diff
@@ -96,6 +96,13 @@ services:
       DATABASE_USER: postgres
       DATABASE_PASSWORD: postgres
       KAFKA_DEMO_TOPIC: demo-topic
+      AWS_ENDPOINT: http://localstack:4566
+      AWS_NO_VERIFY_SSL: "1"
+      AWS_DISABLE_SSL: "1"
+      AWS_SDK_LOAD_CONFIG: "0"
+      AWS_ACCESS_KEY_ID: "test"
+      AWS_SECRET_ACCESS_KEY: "test"
+      AWS_DEFAULT_REGION: "us-east-1"
```
acco: Interesting, are all these env vars actually used?

priv/tests_e2e/tests.exs
```diff
@@ -72,14 +88,54 @@ defmodule Sequin.E2E.KafkaTest do
     end
   end
 
+  describe "sqs integration" do
+    @tag :sqs
+    test "changes are streamed to sqs", %{conn: conn, tid: tid, queue_url: queue_url, aws_client: client} do
+      IO.inspect(queue_url)
```
acco: inspect

Makefile
```diff
@@ -167,6 +167,13 @@ e2e-restart: e2e-down e2e-up ## Restart e2e test environment (shortcut for down+
 e2e-tests: ## Run the e2e test suite
 	cd priv/tests_e2e && elixir tests.exs
 
+e2e-test: ## Run the e2e test suite with specific tag: make e2e-test tag=kafka
+	@if [ -z "$(tag)" ]; then \
+		echo "Usage: make e2e-test tag=<tag-name> (available tags: kafka, sqs)"; \
+		exit 1; \
+	fi
+	cd priv/tests_e2e && elixir tests.exs $(tag)
```
acco: Cool - so this runs a specific, and the one above runs all?

acco (approved):

Makefile
abc3: changed to 
```
make e2e-tests
make e2e-tests sqs
make e2e-tests kafka
```

lib/sequin/consumers/sqs_sink.ex
abc3: @acco could you clarify? these two fields are used to set up localstack

priv/tests_e2e/docker-compose.yml
abc3: removed a few

lib/sequin/consumers/sqs_sink.ex
acco: @abc3 As in instead of `proto` for `http`/`https`, we use `scheme`!

abc3: ahh, gotcha

acco (approved):

PR #1834, opened 2025-06-17
S2 Sink Docs

thisisgoldman:


docs/how-to/stream-postgres-to-s2.mdx
```diff
@@ -0,0 +1,128 @@
...
+
+```bash
+# Configure the CLI with your access token
+s2 config set --access-token <your-access-token>
+
+# Create a basin
+s2 create-basin my-basin
+
+# Create a stream called `products` inside that basin
+s2 create-stream s2://my-basin/products
```
davoclavo, line 50: not critical, but perhaps we could mention that users can create a basin with the `create_stream_on_append` if they want an simpler setup

https://s2.dev/docs/rest/basins/create#body-config-create-stream-on-append

shikhar: Great idea! Including the CLI flag it'd be `s2 create-basin my-basin --create-stream-on-append`

docs/how-to/stream-postgres-to-s2.mdx
```diff
@@ -0,0 +1,128 @@
...
+## Verify & debug
+
+To verify that your S2 sink is working:
+
+1. Make a change in your source table.
+2. Confirm that the **Messages** count for your sink increases in the Sequin console, and observe the new message in the **Messages** tab.
+3. Use the S2 CLI to read from the stream and confirm events arrive:
+
+   ```bash
+   s2 read s2://my-basin/products
```
shikhar: `read` defaults to starting at the tail of the stream, maybe `s2 tail -f s2://my-basin/products` here (which has the regular `tail -f` semantics of also displaying last 10 records and then starting to follow)

docs/reference/sinks/s2.mdx
```diff
@@ -0,0 +1,77 @@
+---
```
shikhar, line 1: Re: the max payload size mentions, the limit is 1 MiB per append

shikhar (approved):
Looks great @thisisgoldman, thank you

docs/how-to/stream-postgres-to-s2.mdx
thisisgoldman: Great suggestion! Adding now

PR #1844, opened 2025-06-18
💄 Surface better error when a publication is in an invalid state

acco:


acco:
* **#1844** <a href="https://app.graphite.dev/github/pr/sequinstream/sequin/1844?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/sequinstream/sequin/1844?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

docs/reference/databases.mdx
```diff
@@ -269,6 +269,34 @@ For Postgres 12 and 13, we do not support backfills, which rely heavily on the `
...
 
+## Known issues
+
+### Publication not recognized by replication slot
+
+When a publication is created *after* the replication slot, Postgres may report that the publication does not exist. Sequin will surface this as an “Issue with publication” health check.
+
+This is a [known Postgres issue](https://www.postgresql.org/message-id/18683-a98f79c0673be358%40postgresql.org).
+
+**Fix**: drop and re-create the replication slot *after* the publication exists:
```
RTLS, line 277: If you want to prevent missing any changes, you should create a new slot, move Sequin to the new slot, then drop the old slot.

lib/sequin/string.ex
```diff
@@ -119,4 +119,10 @@ defmodule Sequin.String do
   def format_bytes(bytes) when bytes < 1024 * 1024, do: "#{Float.round(bytes / 1024, 2)} KB"
   def format_bytes(bytes) when bytes < 1024 * 1024 * 1024, do: "#{Float.round(bytes / 1024 / 1024, 2)} MB"
   def format_bytes(bytes) when bytes < 1024 * 1024 * 1024 * 1024, do: "#{Float.round(bytes / 1024 / 1024 / 1024, 2)} GB"
+
+  def to_existing_atom_safe(string) do
+    String.to_existing_atom(string)
+  rescue
+    _ -> string
+  end
```
RTLS: should this return a status tuple? I feel like this will lead to raises later due to an expected string

RTLS (approved):
Nice

lib/sequin/string.ex
acco: Good point, will change!

docs/reference/databases.mdx
acco: @RTLS The old slot will contain changes that the new slot will not when you create it :( so data loss feels certain. It's a big bummer, as dropping a publication feels non-deleterious, but it can be problematic.

Interestingly, by happenstance, it *might* work as when we connect to the new slot we might send back an old LSN first (cached) and possibly traverse backward - but this would be basically unsupported behavior. We might consider ways to make this operation "first class"

PR #1855, opened 2025-06-20
Add WalSenderMonitor

acco:


lib/sequin/runtime/wal_sender_monitor.ex
```diff
@@ -0,0 +1,120 @@
...
+    Enum.each(previous_states, fn {database_id, prev_wait_event_type, prev_wait_event} ->
+      Gauge.set(
+        [name: @gauge_name, labels: [database_id, prev_wait_event_type, prev_wait_event]],
+        0
+      )
+    end)
+
+    # Set current state to 1
+    Gauge.set(
+      [name: @gauge_name, labels: [db.id, wait_event_type, wait_event]],
```
RTLS, line 111: Did you choose gauge to power a graph? Curious if you considered just logging the query results

lib/sequin/runtime/wal_sender_monitor.ex
```diff
@@ -0,0 +1,120 @@
+defmodule Sequin.Runtime.WalSenderMonitor do
+  @moduledoc false
+  use GenServer, restart: :temporary
```
RTLS, line 3: should we rescue the query so this process is less likely to die and not be restarted?

PR #1864, opened 2025-06-25
✨ Add `metadata.database` object to all messages

acco:


lib/sequin/consumers/consumer_event_data.ex
```diff
@@ -42,6 +42,16 @@ defmodule Sequin.Consumers.ConsumerEventData do
         field :name, :string
         field :annotations, :map
       end
+
+      embeds_one :database, Database, primary_key: false, on_replace: :update do
```
RTLS, line 46: we are not adding this to `ConsumerRecordData` as well?

lib/sequin/consumers/path_function.ex
```diff
@@ -52,6 +52,7 @@ defmodule Sequin.Consumers.PathFunction do
   defp traverse_path(%ConsumerEventData.Metadata{} = data, keys), do: mapify_struct_and_traverse(data, keys)
   defp traverse_path(%ConsumerRecordData.Metadata{} = data, keys), do: mapify_struct_and_traverse(data, keys)
   defp traverse_path(%ConsumerEventData.Metadata.Sink{} = data, keys), do: mapify_struct_and_traverse(data, keys)
+  defp traverse_path(%ConsumerEventData.Metadata.Database{} = data, keys), do: mapify_struct_and_traverse(data, keys)
```
RTLS, line 55: good catch 😅

PR #1874, opened 2025-06-26
fix: improve wal cursor handling logic

davoclavo:


acco:
Handled in: https://github.com/sequinstream/sequin/pull/1896

PR #1902, opened 2025-07-02
✨ AWS SNS Routing

davoclavo:


ADD-SINK-ROUTING.md
```diff
@@ -105,4 +107,17 @@ See `assets/svelte/sinks/gcp_pubsub/GcpPubsubSinkCard.svelte` for an example.
...
 
-Update the reference for the docs to discuss routable fields, ie. `docs/reference/sinks/kafka.mdx`
\ No newline at end of file
+Update the reference for the docs to discuss routable fields, ie. `docs/reference/sinks/kafka.mdx`
+
+### [Backend] Improve Test Connection methods
+
+For sinks that support dynamic routing, improve the testing mechanism so we test credentials or permissions without requiring specific resource access. This is needed because dynamic routing may not have specific topics/queues configured at setup time.
+
+Example: `test_credentials_and_permissions/1` in `lib/sequin/aws/sns.ex`
```
RTLS, line 116: great!

assets/svelte/functions/Edit.svelte
```diff
@@ -853,6 +854,34 @@ Please help me create or modify the Elixir function transform to achieve the des
                     value="nats"
                     label={sinkTypeInternalToExternal.nats}
                   />
+                  <SelectItem
```
RTLS: as noted in discord, we can remove these additions

lib/sequin/runtime/routing/consumers/sns.ex
```diff
@@ -0,0 +1,38 @@
...
+    |> validate_required([:topic_arn])
+    |> validate_length(:topic_arn, min: 1, max: 2000)
+    |> validate_format(:topic_arn, SnsSink.sns_arn_regex(),
+      message: "must be a valid AWS SNS Topic ARN (arn:aws:sns:<region>:<account-id>:<topic-name>)"
+    )
+  end
+
+  def route(_action, _record, _changes, _metadata) do
+    %{
+      topic_arn: nil
```
RTLS, line 29: probably smart to start with nil here 🤔 I did not do this for several of the sinks I added routing for, but I will revist. I think it's best to be clear that this is not set with a system default

lib/sequin/runtime/sns_pipeline.ex
```diff
@@ -27,21 +28,37 @@ defmodule Sequin.Runtime.SnsPipeline do
...
     %{
-      consumer: %SinkConsumer{sink: sink} = consumer,
+      consumer: %SinkConsumer{} = consumer,
       sns_client: sns_client,
       test_pid: test_pid
     } = context
 
     setup_allowances(test_pid)
 
+    routed_messages = Routing.route_and_transform_messages(consumer, messages)
```
RTLS: this will run the user's routing function twice, correct? so we will show `Execution routing function` twice in the Trace tab, for instance.

I think this is fine because it's so cheap to run.

But we can consider only running this in `handle_message` and re-using the result here

lib/sequin/runtime/sns_pipeline.ex
```diff
@@ -50,16 +67,25 @@ defmodule Sequin.Runtime.SnsPipeline do
...
+         transformed_message: transformed_message
+       }) do
+    consumer_data = original_message.data
+
     message = %{
-      message: Sequin.Transforms.Message.to_external(consumer, record_or_event),
-      message_deduplication_id: record_or_event.data.metadata.idempotency_key,
+      message: transformed_message,
+      message_deduplication_id: consumer_data.data.metadata.idempotency_key,
       message_id: UUID.uuid4()
```
RTLS, line 76: so `message_deduplication_id` and `message_id` are not routable but they are included in the routing struct for SNS. We should remove them from the struct presumably?

lib/sequin_web/live/components/consumer_form.ex
```diff
@@ -366,9 +367,24 @@ defmodule SequinWeb.Components.ConsumerForm do
...
+
+        :static ->
+          # Static routing mode - test specific topic access
+          case SNS.topic_meta(client, sink.topic_arn) do
+            :ok -> :ok
+            {:error, error} -> {:error, Exception.message(error)}
+          end
+
+        _ ->
+          {:error, "Invalid routing mode"}
```
RTLS, line 387: 👍

lib/sequin_web/live/components/consumer_form.ex
```diff
@@ -709,17 +725,21 @@ defmodule SequinWeb.Components.ConsumerForm do
       "queue_url" => sink["queue_url"],
       "region" => region,
       "access_key_id" => sink["access_key_id"],
-      "secret_access_key" => sink["secret_access_key"]
+      "secret_access_key" => sink["secret_access_key"],
+      "routing_mode" => sink["routing_mode"]
```
RTLS: these are required to render the consumer form correctly on edit?

thank you I must have missed these 🙏

RTLS (approved):
I don't see edits to the SNS pipeline test but it's probably worth adding a test with routing functions to verify all the new behavior!

lib/sequin/runtime/sns_pipeline.ex
davoclavo: correct - good point, I can adjust this so it only runs once and use it as the batch key, we just have to be careful if we end up adding more fields to the routing struct, that could cause the batching mechanism to change (perhaps that's what we want, I am unsure at the moment)

lib/sequin/runtime/sns_pipeline.ex
davoclavo: thanks, good catch, I added them initially but then removed them so probably forgot to delete them in the routing struct

lib/sequin_web/live/components/consumer_form.ex
davoclavo: let me confirm that

davoclavo: good catch, we don't need to encode it per sink, as it is done at an upper layer:

https://github.com/sequinstream/sequin/blob/main/lib/sequin_web/live/components/consumer_form.ex#L920

lib/sequin/runtime/sns_pipeline.ex
RTLS: yeah this is why I think the routing behaviour needs a callback to specify which keys should be included in the routable batch keys

davoclavo: cool, good idea.

I fixed this for now so it only runs once per message

PR #1904, opened 2025-07-02
Acco/07 01 bug slot producer use last dispatched wal cursor always flush after producing

acco:


lib/sequin/runtime/slot_producer/slot_producer.ex
```diff
@@ -227,7 +227,7 @@ defmodule Sequin.Runtime.SlotProducer do
 
   @impl GenStage
   def handle_demand(incoming, %{demand: demand} = state) do
-    {messages, state} = maybe_produce(%{state | demand: demand + incoming})
+    {messages, state} = maybe_produce_and_flush(%{state | demand: demand + incoming})
```
acco, line 230: This is a critical finding from adding the new test below:

We want to maybe trigger a flush _after every time we produce_.

Before we could get into a situation where a message was in our internal buffer (`accumulated_messages`) and there was no running flush timer. No messages come in from the database. BUT demand comes from downstream. We produce the message, but never trigger the flush to flush it.

lib/sequin/runtime/slot_producer/slot_producer.ex
```diff
@@ -281,8 +281,7 @@ defmodule Sequin.Runtime.SlotProducer do
 
     with {:ok, copies, protocol} <- Protocol.handle_copy_recv(msg, @max_messages_per_protocol_read, protocol),
          {:ok, state} <- handle_copies(copies, %{state | protocol: protocol}) do
-      state = maybe_flush(state)
-      {messages, state} = maybe_produce(state)
+      {messages, state} = maybe_produce_and_flush(state)
```
acco, line 284: Combine so we can't miss critical dependency in future.

lib/sequin/runtime/slot_producer/slot_producer.ex
```diff
@@ -318,26 +317,14 @@ defmodule Sequin.Runtime.SlotProducer do
...
 
   def handle_info(:flush_batch, %State{} = state) do
-    high_watermark_cursor =
-      if state.commit_lsn do
-        # commit_idx starts at 0, and is incremented *after* each assignment. So the last assignment is -1
-        %{commit_lsn: state.commit_lsn, commit_idx: state.commit_idx - 1}
-      else
-        %{commit_lsn: state.last_commit_lsn, commit_idx: state.last_commit_idx}
-      end
+    high_watermark_cursor = state.last_dispatched_wal_cursor || state.restart_wal_cursor
```
acco, line 327: Key change we discussed (great catch!)

lib/sequin/runtime/slot_producer/slot_producer.ex
```diff
@@ -561,14 +548,26 @@ defmodule Sequin.Runtime.SlotProducer do
...
 
+    # Update last_dispatched_wal_cursor to the highest cursor from dispatched messages
+    last_dispatched_wal_cursor =
+      case acc_msgs do
+        [] ->
+          state.last_dispatched_wal_cursor
+
+        msgs ->
+          last_msg = List.last(msgs)
+          %{commit_lsn: last_msg.commit_lsn, commit_idx: last_msg.commit_idx}
```
acco, line 564: Confirm my logic that it's the last msg here

lib/sequin/runtime/slot_producer/slot_producer.ex
```diff
@@ -577,11 +576,11 @@ defmodule Sequin.Runtime.SlotProducer do
...
     }
 
-    {acc_msgs, state}
+    {acc_msgs, maybe_flush(state)}
   end
 
-  defp maybe_produce(state) do
-    {[], state}
+  defp maybe_produce_and_flush(state) do
+    {[], maybe_flush(state)}
```
acco, line 583: Reading this back, we don't need a maybe_flush here?

lib/sequin/runtime/slot_producer/supervisor.ex
```diff
@@ -6,11 +6,6 @@ defmodule Sequin.Runtime.SlotProducer.Supervisor do
   1. SlotProducer - Connects to PostgreSQL replication and produces messages
```
acco, line 6: Flattening stuff

test/sequin/runtime/slot_producer/slot_producer_test.exs
```diff
@@ -383,6 +383,52 @@ defmodule Sequin.Runtime.SlotProducerTest do
       assert SlotProducer.status(slot.id) == :active
     end
 
+    @tag start_opts: [processor_opts: [consumer_demand: :manual]]
+    test "batch marker only reflects dispatched messages, not buffered messages", %{
+      consumer_pid: consumer_pid
+    } do
```
RTLS, line 389: 👍

lib/sequin/runtime/slot_producer/supervisor.ex
```diff
@@ -104,18 +98,13 @@ defmodule Sequin.Runtime.SlotProducer.Supervisor do
...
-    ]
-
-    # Use :rest_for_one so if SlotProducer fails, all downstream stages restart
-    # This ensures proper cleanup and re-subscription of the pipeline
+    children =
+      List.flatten([
+        {SlotProducer, slot_producer_opts},
+        processor_specs,
+        {ReorderBuffer, reoder_buffer_opts}
+      ])
```
RTLS, line 106: 👍 the guarantees of the supervision tree are so much clearer now

lib/sequin/runtime/slot_producer/slot_producer.ex
RTLS: I think this is correct. what does `acc_msgs` stand for? In either case it's not that bad of a bug if you pick the wrong msg from the ones we are producing. The critical bug is to pick a msg from the `remaining` list

lib/sequin/runtime/slot_producer/slot_producer.ex
RTLS: `maybe_flush` depends on `processed_messages_since_last_flush_stats` which is incremented when we handle data and reset when we flush.

does it make more sense to have maybe_flush depend on `produced_messages_since_last_flush_stats` ? flushing is about propagating LSNs of produced messages, right?

wondering if we actually **only** need to trigger `maybe_flush` when we produce messages?

RTLS: I think it's extremely clear if every time we produce messages we simply ensure we have a a flush timer which is always for 10ms in the future. I'm not sure we need the complexity of produced message count / bytes at all?

this works because we are always producing ie. heartbeat messages

lib/sequin/runtime/slot_producer/slot_producer.ex
acco: Just repeating it back to make sure we got it right:

We take accumulated messages, which we've been prepending to:

[ msg3, msg2, msg1 ]

We reverse it, so now it's

[ msg1, msg2, msg3 ]

We split it on demand into {acc_msgs, remaining msgs}. So, if demand is 2, we now have acc_msgs = [ msg1, msg2 ]

So indeed, List.last that and get msg2

lib/sequin/runtime/slot_producer/slot_producer.ex
acco: Agreed!

PR #1906, opened 2025-07-02
Add routing for RabbitMQ sink

RTLS:
## Summary
- add RabbitMQ routing schema and pipeline updates
- update sink config and factories
- support rabbitmq routing in frontend forms and docs
- document routing options for RabbitMQ

------
https://chatgpt.com/codex/tasks/task_e_6864906ebf28832cb60d8470932f52b3

### DIFF
--- a/lib/sequin/runtime/rabbitmq_pipeline.ex
+++ b/lib/sequin/runtime/rabbitmq_pipeline.ex
@@ -3,6 +3,7 @@ defmodule Sequin.Runtime.RabbitMqPipeline do
   @behaviour Sequin.Runtime.SinkPipeline
 
   alias Sequin.Error
+  alias Sequin.Runtime.Routing
   alias Sequin.Runtime.SinkPipeline
   alias Sequin.Sinks.RabbitMq
 
@@ -31,9 +32,9 @@ defmodule Sequin.Runtime.RabbitMqPipeline do
     %{consumer: consumer, test_pid: test_pid} = context
     setup_allowances(test_pid)
 
-    consumer_messages = Enum.map(messages, & &1.data)
+    routed_messages = Routing.route_and_transform_messages(consumer, Enum.map(messages, & &1.data))
 
-    case RabbitMq.send_messages(consumer, consumer_messages) do
+    case RabbitMq.send_messages(consumer, routed_messages) do
       :ok ->
         {:ok, messages, context}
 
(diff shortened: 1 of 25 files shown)


lib/sequin/runtime/rabbitmq_pipeline.ex
```diff
@@ -31,7 +32,10 @@ defmodule Sequin.Runtime.RabbitMqPipeline do
     %{consumer: consumer, test_pid: test_pid} = context
     setup_allowances(test_pid)
 
-    consumer_messages = Enum.map(messages, & &1.data)
+    routing_infos = Routing.route_messages(consumer, Enum.map(messages, & &1.data))
```
davoclavo: ~It seems we are not currently running transform functions for RMQ!~ [nvm, we are executing transforms in client.ex](https://github.com/sequinstream/sequin/blob/davoclavo/feat/add_sns_routing_sink/lib/sequin/sinks/rabbitmq/client.ex#L92-L94)

 perhaps its worth moving that here via `Routing.route_and_transform_messages` and removing that from the client?

lib/sequin/sinks/rabbitmq/client.ex
```diff
@@ -47,15 +47,21 @@ defmodule Sequin.Sinks.RabbitMq.Client do
...
+  defp publish_message(
+         %SinkConsumer{sink: %RabbitMqSink{} = sink} = consumer,
+         message,
+         routing_info,
+         %AMQP.Channel{} = channel
+       ) do
     # https://hexdocs.pm/amqp/AMQP.Basic.html#publish/5-options
-    # FIXME: message.id should not be relied on?
-    opts = [message_id: to_string(message.id), content_type: "application/json"]
+    exchange = routing_info.exchange || sink.exchange
```
davoclavo: I think this default is already covered by `def route_consumer`? unless you are thinking of also adding a default for dynamic routing functions?

lib/sequin/runtime/rabbitmq_pipeline.ex
RTLS: great catch! this is a good cleanup

lib/sequin/sinks/rabbitmq/client.ex
RTLS: yes good catch!

PR #1907, opened 2025-07-02
Updates to SlotProducer from code review

acco:


lib/sequin/runtime/slot_producer/processor.ex
```diff
@@ -130,13 +130,6 @@ defmodule Sequin.Runtime.SlotProducer.Processor do
         case Decoder.decode_message(msg.payload) do
           %type{} = payload when type in [Insert, Update, Delete, LogicalMessage] ->
             %{msg | message: cast_message(payload, msg, state.relations)}
-
-          %type{} = _msg when type in [Origin, Truncate, Type] ->
-            nil
-
-          msg ->
-            Logger.warning("Unhandled message type: #{inspect(msg)}")
-            nil
```
acco, line 139: Safe to do as SlotProducer filters everything

lib/sequin/runtime/slot_producer/slot_producer.ex
```diff
@@ -121,38 +121,37 @@ defmodule Sequin.Runtime.SlotProducer do
       # Cursors & aliveness
       field :restart_wal_cursor, Replication.wal_cursor()
       field :restart_wal_cursor_fn, (PostgresReplicationSlot.id(), Postgres.wal_cursor() -> Postgres.wal_cursor())
-      field :setting_update_cursor_interval, non_neg_integer()
+      field :setting_restart_wal_cursor_update_interval, non_neg_integer()
+      field :restart_wal_cursor_update_timer, reference()
       field :setting_ack_interval, non_neg_integer()
       field :ack_timer, reference()
-      field :update_cursor_timer, reference()
       field :last_dispatched_wal_cursor, Replication.wal_cursor()
```
acco, line 128: rename + sort

lib/sequin/runtime/slot_producer/slot_producer.ex
```diff
@@ -121,38 +121,37 @@ defmodule Sequin.Runtime.SlotProducer do
...
 
       # Batches
       field :batch_idx, BatchMarker.idx(), default: 0
-      field :last_flushed_batch_at, DateTime.t()
 
-      field :processed_messages_since_last_flush_stats, %{count: non_neg_integer(), bytes: non_neg_integer()},
-        default: %{count: 0, bytes: 0}
+      field :processed_messages_since_last_flush_stats,
+            %{count: non_neg_integer(), bytes: non_neg_integer(), last_flushed_at: DateTime.t()},
+            default: %{count: 0, bytes: 0, last_flushed_at: nil}
```
acco, line 135: consolidate

lib/sequin/runtime/slot_producer/slot_producer.ex
```diff
@@ -121,38 +121,37 @@ defmodule Sequin.Runtime.SlotProducer do
...
 
       field :batch_flush_timer, {reference(), non_neg_integer()}
       field :setting_batch_flush_interval, SlotProducer.batch_flush_interval()
 
       # Current xaction state
       field :commit_ts, DateTime.t()
       field :commit_lsn, integer()
-      field :commit_idx, integer()
+      field :next_commit_idx, integer()
```
acco, line 143: rename (functionality/value stays the same)

lib/sequin/runtime/slot_producer/slot_producer.ex
```diff
@@ -211,8 +210,9 @@ defmodule Sequin.Runtime.SlotProducer do
       restart_wal_cursor_fn: Keyword.fetch!(opts, :restart_wal_cursor_fn),
       setting_reconnect_interval: Keyword.get(opts, :reconnect_interval, :timer.seconds(10)),
       setting_ack_interval: Keyword.get(opts, :ack_interval, :timer.seconds(10)),
-      setting_update_cursor_interval: Keyword.get(opts, :update_cursor_interval, :timer.seconds(10)),
-      processor_mod: Keyword.get(opts, :processor_mod),
+      setting_restart_wal_cursor_update_interval:
+        Keyword.get(opts, :restart_wal_cursor_update_interval, :timer.seconds(10)),
```
acco, line 214: rename to match

lib/sequin/runtime/slot_producer/slot_producer.ex
```diff
@@ -353,17 +352,22 @@ defmodule Sequin.Runtime.SlotProducer do
...
-          if not is_nil(state.commit_lsn) and next_state.commit_lsn == state.commit_lsn do
-            {:cont, {:ok, %{next_state | commit_idx: state.commit_idx + 1}}}
+        {:ok, type, next_state} ->
+          advance_next_commit_idx? = not is_nil(state.commit_lsn) and type in @advance_idx_types
+
+          if advance_next_commit_idx? do
+            {:cont, {:ok, %{next_state | next_commit_idx: state.next_commit_idx + 1}}}
           else
             {:cont, {:ok, next_state}}
           end
```
acco: I decided it was safest to have `handle_data` return the type, and then do next_commit_idx advancement up here. So critical, felt it made sense to do in one place vs many.

lib/sequin/runtime/slot_producer/slot_producer.ex
```diff
@@ -353,17 +352,24 @@ defmodule Sequin.Runtime.SlotProducer do
...
         {:ok, next_state} ->
-          if not is_nil(state.commit_lsn) and next_state.commit_lsn == state.commit_lsn do
-            {:cont, {:ok, %{next_state | commit_idx: state.commit_idx + 1}}}
+          advance_next_commit_idx? = not is_nil(state.commit_lsn) and type in @advance_idx_types
+
+          if advance_next_commit_idx? do
+            {:cont, {:ok, %{next_state | next_commit_idx: state.next_commit_idx + 1}}}
           else
             {:cont, {:ok, next_state}}
           end
```
acco, line 375: Couple big changes here:

- We have the type up here, so we can determine if we should advance idx (I decided it was so important we should do in one place vs in each handler)
- To get the type, we do a `parse_copy` and then later `handle_data` with the type extracted. According to some brief research, this is probably better than what we were doing before, which was popping the type/head off the binary, and then adding back on (for stuff like the decoder)

I figured it was actually nicer to not have the duplicitous `<<?w, _header::192, ` in all func headers too

RTLS: this all makes sense to me!

RTLS:
```
iex(54)> {time, _} = :timer.tc(fn -> <<"a", binary::binary>> end); time / 1000
0.028
iex(55)> {time, _} = :timer.tc(fn -> binary end); time / 1000
0.005
```

this might be much faster!?

test/sequin/postgres_replication_test.exs
```diff
@@ -971,9 +971,9 @@ defmodule Sequin.PostgresReplicationTest do
       [insert1, insert2, insert3] = changes
 
       # Assert seq values increase within transaction
-      assert insert1.commit_idx == 1
-      assert insert2.commit_idx == 2
-      assert insert3.commit_idx == 3
+      assert insert1.commit_idx == 0
+      assert insert2.commit_idx == 1
+      assert insert3.commit_idx == 2
```
RTLS, line 976: 💆🏼‍♂️

RTLS (approved):

PR #1919, opened 2025-07-02
✨ Improve functions auto complete

RTLS:


lib/sequin_web/live/functions/auto_complete.ex
```diff
@@ -1,5 +1,6 @@
 defmodule SequinWeb.FunctionLive.AutoComplete do
   @moduledoc false
+  @reserved_words ~w(do if else end)
```
davoclavo: I would suggest adding `case` as well - possibly `cond`, `with`, `when` and `for` too?

Issue #1800, opened 2025-06-12
We should compress external outbound payloads


davoclavo:
Regarding HTTP:
 - The easiest approach is to use gzip with Req via [`compress_body: true`](https://github.com/wojtekmach/req/blob/c17cd200130e985b06b25091da07ffb5fc95c9cc/lib/req/steps.ex#L699-L716)
 - The best approach is to use faster compression mechanisms via our own custom pipeline that compresses before passing it to Req, such as zstd or brotli. Make sure proper headers are added.

For other protocols:
 - Investigate what options each client has
 - Fallback to compress on or own and pass compressed payloads. Make sure proper headers/metadata is added.

For either case, we should make sure that the receiving end is able to receive the compressed payloads. And also it would be good to run some benchmarks to understand the tradeoffs between latency/throughput and CPU usage

PR #1932, opened 2025-07-03
⚡️ Add outbound gzip compression for GCP Pubsub and Elasticsearch sinks

davoclavo:
### Add outbound gzip compression for GCP Pubsub and Elasticsearch sinks

  
  **Currently Enabled (2 sinks)**

  | Sink        | Library | Compression Status | Implementation                      |
  |-------------|---------|--------------------|-------------------------------------|
  | S2          | Req     | ✅ Enabled          | compress_body: true in base_request |
  | Meilisearch | Req     | ✅ Enabled          | compress_body: true in base_request |

  **Enabled in this PR (2 sinks)**

  | Sink            | Library | Compression Status | Implementation Path                              |
  |-----------------|---------|--------------------|--------------------------------------------------|
  | GCP Pub/Sub     | Req     | 🟢 Ready           | Add compress_body: true |
  | Elasticsearch   | Req     | 🟢 Ready           | Add compress_body: true |

  **Enabled in this PR (2 sinks)**
  | Sink            | Library | Compression Status | Implementation Path                              |
  |-----------------|---------|--------------------|--------------------------------------------------|
  | HTTP Push       | Req     | 🟡 WIP           |  See #1934 |
  | Kafka          | :brod   | 🟡 WIP           | Set the right `:brod` options configuration|

  
   **Discovered these services do not support transparent compression (5 sinks)**

  | Sink            | Compression Status | Reference              |
  |-----------------|--------------------|------------------------|
  | Azure Event Hub | 🟠 Needs custom consumers   |  |
  | NATS            | 🟠 Needs custom consumers   |  |
  | Redis String    | 🟠 Needs custom consumers   |  |
  | Redis Stream   | 🟠 Needs custom consumers   |  |
  | Kinesis         | 🟠 Needs custom consumers   |                        |
  | SNS             | 🟠 Needs custom consumers   |                        |
  | SQS             | 🟠 Needs custom consumers   |                        |
  | Typesense       | 🔴 Not supported yet  | https://github.com/typesense/typesense/issues/580                        |

This PR improves the situation for #1800 but doesn't close the issue until we figure out what to do with the pending sinks

### Additional thoughts
  - [x] We need to confirm that all the receiving ends support gzip compressed payloads, but I would be surprised if any of them did not supported them
  - [ ] In the future we could consider supporting other compression mechanisms, which may be better suited than gzip, such as: snappy, brotli, lz4, zstd. But we would need to check compatibility with the receiving end
  - [ ] We could also consider in the future having a compression algorithm attribute in the routing struct, or a setting sink consumers, so users can pick their compression mechanism

---

History log (before service review where we discarded incompatible services)

0e5f476b88a782c71f2dc9998fa13d2efa8db79e

### DIFF
--- a/lib/sequin/runtime/elasticsearch_pipeline.ex
+++ b/lib/sequin/runtime/elasticsearch_pipeline.ex
@@ -13,8 +13,9 @@ defmodule Sequin.Runtime.ElasticsearchPipeline do
   require Logger
 
   @impl SinkPipeline
-  def init(context, _opts) do
-    context
+  def init(context, opts) do
+    req_opts = Keyword.get(opts, :req_opts, [])
+    Map.put(context, :req_opts, req_opts)
   end
 
   @impl SinkPipeline
@@ -45,7 +46,7 @@ defmodule Sequin.Runtime.ElasticsearchPipeline do
 
   @impl SinkPipeline
   def handle_batch(:default, messages, batch_info, context) do
-    %{consumer: consumer, test_pid: test_pid} = context
+    %{consumer: consumer, test_pid: test_pid, req_opts: req_opts} = context
     index_name = batch_info.batch_key
     sink = consumer.sink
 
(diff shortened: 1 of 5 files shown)


RTLS (approved):
thanks for adding elasticsearch pipeline tests 🙌

Issue #1935, opened 2025-07-06
Do delete actions require any additional setup for Typesense sink?


alisamar:
I've set up a Typesense sink and things are mostly working great. However, I'm running into an issue where insert and update actions succeed, but **delete** actions consistently fail with the following error:
```
[sink_pipeline]: Missing document ID for deletion
```

The transform function I’m using includes the id field like this:

```elixir
def transform(action, record, changes, metadata) do
  case action do
    "delete" ->
      %{
        id: record["id"],
      }

    _ ->
      %{
        id: record["id"],
        name: record["name"],
        ...
      }
  end
end
```

And the incoming message for delete includes the `record["id"]` field correctly. Despite that, the message fails with the `"Missing document ID"` error and gets stuck in a backoff loop.

Is there any extra configuration or expectation specifically for delete actions when using the Typesense sink? Should I be using `metadata.record_pks` instead?

Thanks for your help!

**Typesense**: v29.0
**Sequin**: v0.12.1
**PostgreSQL**: v16

> **Messages**
> 
> ```
> Seq: 122067528
> State: backing off
> PKs: 1f6a84e0-3f04-4c43-92c6-cc684e139cf9
> Deliver Count: 7
> Last Delivered At: Today at 11:47:57 AM
> Backoff Until: Today at 11:48:57 AM
> Committed At: Today at 11:46:19 AM
> ```
> 
> **Message Details**
> 
> ```
> State: backing off
> Record PKs: 1f6a84e0-3f04-4c43-92c6-cc684e139cf9
> Commit LSN: 122067528
> Deliver Count: 8
> Last Delivered: Today at 11:48:57 AM
> Not visible until: Today at 11:51:19 AM
> ```
> 
> **Message shape**
> 
> ```json
> {
>   "record": {
>     "created_at": "2025-07-06T08:45:42.325203Z",
>     "id": "1f6a84e0-3f04-4c43-92c6-cc684e139cf9",
>     "name": "Demo 1",
>     "updated_at": "2025-07-06T08:45:42.325203Z",
>     "user_id": "demo1"
>   },
>   "metadata": {
>     "commit_lsn": 122067528,
>     "commit_idx": 0,
>     "record_pks": [
>       "1f6a84e0-3f04-4c43-92c6-cc684e139cf9"
>     ],
>     "commit_timestamp": "2025-07-06T08:46:19.237948Z",
>     "table_schema": "public",
>     "table_name": "test",
>     "database_name": "prod-db",
[...]

PR #1936, opened 2025-07-07
Extract document_id from "id" or :id for typesense delete_document

davoclavo:
Accept `:id` as a valid key to extract document id for typesense delete_document action

See #1935

### DIFF
--- a/test/sequin/typesense_pipeline_test.exs
+++ b/test/sequin/typesense_pipeline_test.exs
@@ -135,6 +135,41 @@ defmodule Sequin.Runtime.TypesensePipelineTest do
       assert_receive {:ack, _ref, successful, []}, 3000
       assert length(successful) == 3
     end
+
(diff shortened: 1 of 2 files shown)


test/sequin/typesense_pipeline_test.exs
```diff
@@ -135,6 +135,41 @@ defmodule Sequin.Runtime.TypesensePipelineTest do
...
+
+      start_pipeline!(consumer, adapter)
+
+      send_test_batch(consumer, [message])
+
+      assert_receive {:typesense_request, _sink}, 3000
+
+      assert_receive {:ack, _ref, [success], []}, 3000
+      assert success.data == message
+    end
```
RTLS: shouldn't this test use an atom for `:id`? I believe most accurately it should use a routing function to return `%{id: record["id"]}`

davoclavo: you are right! I had it as an atom before and I moved some code and accidentally reverted the changes

PR #1939, opened 2025-07-07
🔧 Fix flaky connection_cache_tests

davoclavo:
Fix some of the tests that often fail during signoff.

1. Appending `_changed` to the database name factory results forces the config to be different, rather than relying on non-deterministic randomness
2. Adding a 1ms sleep addresses a race condition - I couldn't find the root cause but this seems to fix it for now.

Validated running the test file many times - without these changes the tests often fail at the 10th or so attempt

`mix test test/sequin/connection_cache_test.exs --repeat-until-failure 1000`


acco:
Thanks @davoclavo !

PR #1946, opened 2025-07-08
✨ Load AWS credentials using task role, env, or profile

### LINKED ISSUE #1341: AWS credentials loading
I would like to deploy Sequin in AWS ECS. When using AWS sinks (e.g. SQS), how can I leverage my ECS task role? I don't want to provision an IAM user, but rather use the credentials available in the compute environment in which Sequin is running.

As far as I can tell, the AWS sinks all require AWS creds to be provided as part of the configuration. Are AWS SDKs being used under the hood? AWS SDKs automatically find the credentials available to the ECS task (associated with the task role). If these could be omitted, and discovered using the AWS SDK credentials loading, that would simplify infrastructure-as-code deployments.

davoclavo:
Load AWS-related credentials from Task Role, ENV variables or profile using https://github.com/aws-beam/aws_credentials

From their README:

> Retrieve AWS credentials from a variety of possible sources in the following default order:
> 1. Erlang application environment variables
> 2. OS environment variables
> 3. An AWS [credentials file][1]
> 4. ECS task credentials
> 5. EC2 metadata

### UI changes:
Sink edit form
image

Sink show card
image

### Error states:

Invalid permissions:
image

No credentials available:
image

---

Fixes #1341


assets/svelte/components/AwsAuthenticationForm.svelte
```diff
@@ -0,0 +1,75 @@
+<script lang="ts">
```
RTLS, line 1: nice component here 👍

docs/reference/sinks/sqs.mdx
```diff
@@ -53,6 +59,51 @@ Sequin requires an IAM user with the following permissions:
...
+
+### Using task roles (recommended for AWS environments)
+
+When deploying Sequin on AWS infrastructure, you can attach the above policy directly to:
+
+- **ECS Task Role**: For ECS deployments, attach the policy to the task role
+- **EC2 Instance Profile**: For EC2 deployments, attach the policy to the instance profile  
+- **EKS Service Account**: For EKS deployments, use IAM Roles for Service Accounts (IRSA)
+
+Then set `use_task_role: true` in your sink configuration. Sequin will automatically discover and use the credentials from the environment.
```
RTLS: do we need similar docs for SNS and Kinesis? Note you can use mintlify snippets to share some content between different files

RTLS: these docs are great 🔥

docs/reference/sinks/sqs.mdx
```diff
@@ -53,6 +59,51 @@ Sequin requires an IAM user with the following permissions:
...
+### Using task role (recommended for AWS environments)
+
+```yaml
+sinks:
+  - name: "sqs-sink"
+    batch_size: 10
+    destination:
+      type: "sqs"
+      queue_url: "https://sqs.us-east-1.amazonaws.com/123456789012/my-queue"
+      use_task_role: true
```
RTLS, line 76: we probably need to update the `sequin-yaml.mdx` and the management API docs too!

lib/sequin/aws/task_role_credentials.ex
```diff
@@ -0,0 +1,92 @@
...
+
+  @doc """
+  Gets AWS credentials from the environment using the aws_credentials library.
+
+  Returns `{:ok, credentials}` on success or `{:error, reason}` on failure.
+  """
+  @spec get_credentials() :: {:ok, map()} | {:error, term()}
+  def get_credentials do
+    case Application.ensure_all_started(:aws_credentials) do
+      {:ok, _} ->
```
RTLS: interesting -- is this necessary?

I see in the [aws_credential docs](https://github.com/aws-beam/aws_credentials) that it specifies this call. I am hopeful though that mix will start this automatically in the Elixir project runtime

lib/sequin/aws/task_role_credentials.ex
```diff
@@ -0,0 +1,92 @@
...
+        case :aws_credentials.get_credentials() do
+          :undefined ->
+            {:error, "No credentials available"}
+
+          credentials when is_map(credentials) ->
+            {:ok, credentials}
+
+          other ->
+            {:error, "Unexpected credential format: #{inspect(other)}"}
```
RTLS: I might recommend changing the spec to:
```
| {:error, Error.t()}
```

and then constructing ie. `Sequin.Error.invariant/1` here

Ditto below you can wrap the returned reason in a `Sequin.Error`

lib/sequin/consumers/kinesis_sink.ex
```diff
@@ -18,18 +18,31 @@ defmodule Sequin.Consumers.KinesisSink do
...
     |> validate_stream_arn()
     |> validate_routing()
   end
 
+  defp validate_credentials(changeset) do
+    use_task_role = get_field(changeset, :use_task_role)
+
+    if use_task_role do
+      # When using task role, we only need region
+      validate_required(changeset, [:region])
```
RTLS: optional: you can `put_change(:access_key_id, nil)` etc. here to wipe out these fields when using task role. up to you

lib/sequin/runtime/kinesis_pipeline.ex
```diff
@@ -3,15 +3,25 @@ defmodule Sequin.Runtime.KinesisPipeline do
...
+    %{consumer: consumer, test_pid: test_pid} = context
+
+    setup_allowances(test_pid)
+
+    case KinesisSink.aws_client(consumer.sink) do
+      {:ok, client} ->
+        Map.put(context, :kinesis_client, client)
+
+      {:error, reason} ->
+        raise "Failed to initialize Kinesis client: #{inspect(reason)}"
```
RTLS, line 23: 👍 might as well blow up here

mix.lock
```diff
@@ -45,9 +47,7 @@
...
   "ex_aws_sqs": {:hex, :ex_aws_sqs, "3.4.0", "f7c4d0177c1c954776363d3dc05e5dfd37ddf0e2c65ec3f047e5c9c7dd1b71ac", [:mix], [{:ex_aws, "~> 2.1", [hex: :ex_aws, repo: "hexpm", optional: false]}, {:hackney, "~> 1.9", [hex: :hackney, repo: "hexpm", optional: true]}, {:jason, "~> 1.1", [hex: :jason, repo: "hexpm", optional: true]}, {:saxy, "~> 1.1", [hex: :saxy, repo: "hexpm", optional: true]}, {:sweet_xml, ">= 0.0.0", [hex: :sweet_xml, repo: "hexpm", optional: true]}], "hexpm", "b504482206ccaf767b714888e9d41a1cfcdcb241577985517114191c812f155a"},
-  "ex_edn": {:hex, :ex_edn, "0.1.2", "50beb2909a6614a7f5a42296701138bff2c3301949181455e7da4cc997a95ac5", [:mix], [{:array, "~> 1.0.1", [hex: :array, repo: "hexpm", optional: false]}, {:timex, "~> 0.13.4", [hex: :timex, repo: "hexpm", optional: false]}], "hexpm", "9568e79cb96bd61f26389b96ab1ac4fee57762fba6c166ddc745c521ccf2c5ca"},
```
RTLS, line 49: oops thanks!

test/sequin/consumers/sns_sink_test.exs
```diff
@@ -0,0 +1,172 @@
+defmodule Sequin.Consumers.SnsSinkTest do
```
RTLS, line 1: Thank you for creating test file here 🙏

test/sequin/kinesis_pipeline_test.exs
```diff
@@ -14,6 +15,11 @@ defmodule Sequin.Runtime.KinesisPipelineTest do
   alias Sequin.Functions.MiniElixir
   alias Sequin.Runtime.SinkPipeline
 
+  setup do
+    Mox.verify_on_exit!()
```
RTLS: I think this call is not necessary - we do `Hammox.verify_on_exit!` in `Sequin.Case`

test/sequin/kinesis_pipeline_test.exs
```diff
@@ -224,6 +230,79 @@ defmodule Sequin.Runtime.KinesisPipelineTest do
       # Verify we got exactly 2 requests
       assert :counters.get(request_count, 1) == 2
     end
+
+    test "events are sent to Kinesis with use_task_role=true", %{consumer: consumer} do
+      test_pid = self()
+
+      # Configure application to use the mock
+      Application.put_env(:sequin, :task_role_credentials_module, TaskRoleCredentialsMock)
```
RTLS: This is not safe with `async: true` tests. Application env is global, so if this value has to differ between test files then we have a race condition.

I think you set this already in `config/test.exs` though so perhaps we just don't need this line?

test/sequin/kinesis_pipeline_test.exs
```diff
@@ -224,6 +230,79 @@ defmodule Sequin.Runtime.KinesisPipelineTest do
...
+         "task-role-access-key"
+         |> AWS.Client.create("task-role-secret-key", "us-east-1")
+         |> Map.put(:session_token, "task-role-session-token")}
+      end)
+
+      on_exit(fn ->
+        Application.delete_env(:sequin, :task_role_credentials_module)
+        # Reset mox mode for this mock only
+        Mox.set_mox_private(TaskRoleCredentialsMock)
+      end)
```
RTLS: I don't understand why this is necessary 🤔

test/sequin/sqs_sink_test.exs
```diff
@@ -1,8 +1,13 @@
 defmodule Sequin.Consumers.SqsSinkTest do
   use ExUnit.Case, async: true
 
+  import Mox
+  import Sequin.Test.AwsTestHelpers
+
   alias Sequin.Consumers.SqsSink
 
+  setup :verify_on_exit!
```
RTLS: ditto

test/sequin_web/controllers/yaml_controller_test.exs
```diff
@@ -638,5 +638,161 @@ defmodule SequinWeb.YamlControllerTest do
...
+      refute Map.has_key?(sqs_sink["destination"], "secret_access_key")
+
+      sns_sink = Enum.find(parsed_yaml["sinks"], fn sink -> sink["name"] == "sns-task-role-sink" end)
+      assert sns_sink["destination"]["use_task_role"] == true
+      refute Map.has_key?(sns_sink["destination"], "access_key_id")
+      refute Map.has_key?(sns_sink["destination"], "secret_access_key")
+
+      kinesis_sink = Enum.find(parsed_yaml["sinks"], fn sink -> sink["name"] == "kinesis-task-role-sink" end)
+      assert kinesis_sink["destination"]["use_task_role"] == true
+      refute Map.has_key?(kinesis_sink["destination"], "access_key_id")
```
RTLS, line 700: awesome

docs/reference/sinks/sqs.mdx
acco: Yeah, great job!

lib/sequin/aws/task_role_credentials.ex
acco: And if it doesn't, I'm sure we can do this in application.ex

lib/sequin/aws/task_role_credentials.ex
```diff
@@ -0,0 +1,92 @@
...
+    end
+  end
+
+  @doc """
+  Gets AWS credentials from the environment using the aws_credentials library.
+
+  Returns `{:ok, credentials}` on success or `{:error, reason}` on failure.
+  """
+  @spec get_credentials() :: {:ok, map()} | {:error, term()}
+  def get_credentials do
```
acco: We should block this API at multiple layers if self_hosted=false

So we could do a check here and raise if we try to hit this path when self_hosted=false

lib/sequin/consumers/kinesis_sink.ex
acco: I think this is a good idea! Otherwise, it can be surprising weeks later to edit a sink only to discover it's been "hiding" a secret the whole time

lib/sequin/runtime/sns_pipeline.ex
```diff
@@ -88,5 +98,6 @@ defmodule Sequin.Runtime.SnsPipeline do
   defp setup_allowances(test_pid) do
     Req.Test.allow(Sequin.Aws.HttpClient, test_pid, self())
     Mox.allow(Sequin.TestSupport.DateTimeMock, test_pid, self())
+    Mox.allow(TaskRoleCredentialsMock, test_pid, self())
```
acco: Way to figure out this pattern!

test/sequin/aws/task_role_credentials_test.exs
```diff
@@ -0,0 +1,97 @@
+defmodule Sequin.Aws.TaskRoleCredentialsTest do
+  use ExUnit.Case, async: true
+
+  import Mox
+  import Sequin.Test.AwsTestHelpers
+
+  alias Sequin.Aws.TaskRoleCredentials
```
acco: Use `Sequin.Case`, and you'll get Hammox (just a layer up on top of Mox) and :verify_on_exit! imported for you

test/sequin/aws/task_role_credentials_test.exs
```diff
@@ -0,0 +1,97 @@
...
+      }
+
+      client = TaskRoleCredentials.build_client(credentials, "us-west-2")
+
+      assert client.access_key_id == "ASIA123456789"
+      assert client.secret_access_key == "test_secret"
+      assert client.region == "us-west-2"
+      assert client.session_token == "session_token"
+    end
+  end
```
acco: These tests are marginal and I wouldn't be sad to see them go!

test/sequin/aws/task_role_credentials_test.exs
```diff
@@ -0,0 +1,97 @@
...
+
+  describe "using TestCredentials with environment variables" do
+    test "get_client/1 returns client when credentials are set via environment" do
+      setup_aws_credentials(
```
acco: We don't use a lot of imports - I'd just call `AwsTestHelpers.setup_aws_credentials` - reads just fine

test/sequin/consumers/sns_sink_test.exs
```diff
@@ -0,0 +1,172 @@
...
+  use ExUnit.Case, async: true
```
acco: Sequin.Case

test/sequin/consumers/sns_sink_test.exs
```diff
@@ -0,0 +1,172 @@
...
+  setup do
+    # Configure application to use the mock
+    Application.put_env(:sequin, :task_role_credentials_module, TaskRoleCredentialsMock)
+
+    on_exit(fn ->
+      Application.delete_env(:sequin, :task_role_credentials_module)
+    end)
+
+    :ok
+  end
```
acco: The problem with this route is that you're affecting global state, so would need to do `async: false`.

Preferred pattern here: you can do that we do elsewhere you can see with `Sequin` modules, like `Sequin.uuid`:

- Have the module in test *always* be the mock (static) (i.e. hard-code in `test.exs`)
- In [`default_stubs`](https://github.com/sequinstream/sequin/blob/42dbc6868f6dd20d41de15335f614b57865814dc/test/support/case.ex#L40-L48), stub to use the real thing
- Then in test, overwrite the default stub as needed with the mock behavior

test/sequin/sqs_pipeline_test.exs
```diff
@@ -13,6 +15,11 @@ defmodule Sequin.Runtime.SqsPipelineTest do
   alias Sequin.Runtime.SlotMessageStore
   alias Sequin.Runtime.SlotMessageStoreSupervisor
 
+  setup do
+    Mox.verify_on_exit!()
+    :ok
+  end
```
acco: ditto

lib/sequin/aws/task_role_credentials.ex
davoclavo: I believe we do not want to start the application automatically, because this app will perform HTTP requests when its enabled (in order to fetch the task role and creds from the AWS service)

Let me confirm if enabling it but not using it does not cause the checks to be executed.

lib/sequin/aws/task_role_credentials.ex
```diff
@@ -0,0 +1,92 @@
+defmodule Sequin.Aws.TaskRoleCredentials do
```
acco: This could maybe just live as functions under `Sequin.Aws`. And then we make that the mockable thing.

See comment in the file test/sequin/consumers/sns_sink_test.exs below

test/sequin/kinesis_pipeline_test.exs
davoclavo: @acco @RTLS There is an issue when mixing Req + Hammox/Mox `verify_on_exit!` seems to be causing some `expect` not to be verified - I can am working on a fix, potentially a PR to Req's library. Here is a quick summary:

https://github.com/wojtekmach/req/issues/490

test/support/aws_test_helpers.ex
```diff
@@ -0,0 +1,139 @@
...
+      setup do
+        setup_aws_credentials(
+          access_key_id: "AKIATEST123456789",
+          secret_access_key: "my_test_secret",
+          session_token: "test_token",
+          region: "us-west-2"
+        )
+      end
+  """
+  def setup_aws_credentials(opts \\ []) do
```
davoclavo: Consider removing this method of injecting ENVs + running aws_credentials app in tests

1) To avoid polluting test environment
2) To avoid the misleading background log error msgs

```
[error] [17:21:36.018] Provider :aws_credentials_env reports {:error, :environment_credentials_unavailable} line=202 pid=<0.347.0> 
...
```

lib/sequin/runtime/kinesis_pipeline.ex
davoclavo: just to confirm, are you saying that it's ok to raise here? or should we return an `{:error, err}` tuple?

lib/sequin/aws/task_role_credentials.ex
davoclavo: Regarding this.. @RTLS you were right, the application was started automatically by the Elixir runtime. That said, I believe that we do not want to automatically start the `:aws_credentials` application because it will start attempting contacting a credential AWS server, and constantly print some error logs to the console in case it is unable to do so.

So I believe turning it on _only_ if its being used its the best way forward. In order to accomplish this I also made a change in `mix.exs` to turn it off during boot time `{:aws_credentials, "~> 0.3.4", runtime: false}`

test/support/case.ex
```diff
@@ -65,4 +66,18 @@ defmodule Sequin.Case do
...
   end
+
+  def setup_self_hosted_tag(context) do
+    if Map.has_key?(context, :self_hosted) do
+      self_hosted = Map.get(context, :self_hosted)
+      original_self_hosted = Application.get_env(:sequin, :self_hosted)
+      Application.put_env(:sequin, :self_hosted, self_hosted)
+
+      on_exit(fn ->
+        Application.put_env(:sequin, :self_hosted, original_self_hosted)
```
davoclavo: @RTLS @acco This is the last place where calls to `Application.put_env` are done, which as you mentioned affect globally. I could wrap this in a module so we can mock it, but I have ran the tests multiple times without issues.

Let me know if you think its a better approach to do so, specifically for this flag.

Relevant comments:

https://github.com/sequinstream/sequin/pull/1946#discussion_r2198624059
https://github.com/sequinstream/sequin/pull/1946#discussion_r2198725846

davoclavo: Update: I am working on a mock for this to avoid any issues in the future

lib/sequin/runtime/kinesis_pipeline.ex
RTLS: I think raising here is the correct behavior

lib/sequin/aws/task_role_credentials.ex
davoclavo: Added validations for self_hosted in the changeset validations, and in the public functions of the aws client creation.

lib/sequin/aws/task_role_credentials.ex
RTLS: okay this makes sense! sort of unfortunate that the library automatically makes requests in the background and logs errors. but given that behavior I think this is a good solution 👍

lib/sequin/config.ex
```diff
@@ -0,0 +1,34 @@
+defmodule Sequin.Config do
```
RTLS, line 1: 👍

test/support/aws_test_helpers.ex
```diff
@@ -0,0 +1,72 @@
...
+    session_token = Keyword.get(opts, :session_token)
+    region = Keyword.get(opts, :region, "us-east-1")
+
+    Mox.stub(AwsMock, :get_client, fn _region ->
+      client =
+        access_key_id
+        |> AWS.Client.create(secret_access_key, region)
+        |> then(fn client ->
+          if session_token, do: Map.put(client, :session_token, session_token), else: client
+        end)
```
RTLS: potentially `Sequin.Map.put_if_present` would be helpful here

RTLS (approved):

PR #1960, opened 2025-07-09
✨ Generate proper group_id, idempotency key in backfills

acco:


lib/sequin/consumers/consumers.ex
```diff
@@ -792,8 +792,20 @@ defmodule Sequin.Consumers do
...
+    metadata = metadata(message, consumer, database)
+    metadata = struct(ConsumerEventData.Metadata, metadata)
+
+    %ConsumerEventData{
+      record: message_record(message),
+      changes: nil,
+      action: :read,
+      metadata: metadata
+    }
+  end
```
acco: Moved TableReader over to using the same ConsumerEvent/Record generation code that MessageHandler/rest of the pipeline does

lib/sequin/consumers/consumers.ex
```diff
@@ -756,14 +756,15 @@ defmodule Sequin.Consumers do
...
     metadata = struct(ConsumerEventData.Metadata, metadata)
 
     %ConsumerEventData{
       record: message_record(message),
       changes: nil,
-      action: :insert,
+      action: action,
       metadata: metadata
     }
   end
```
acco, line 770: Moved TableReader over to using the same ConsumerEvent/Record generation code that MessageHandler/rest of the pipeline does

Combined these function clauses as the are same

lib/sequin/consumers/consumers.ex
```diff
@@ -850,7 +839,7 @@ defmodule Sequin.Consumers do
       commit_idx: message.commit_idx,
       consumer: consumer_metadata(consumer),
       transaction_annotations: nil,
-      idempotency_key: Base.encode64("#{message.commit_lsn}:#{message.commit_idx}"),
+      idempotency_key: message.idempotency_key,
```
acco, line 842: Rely on idempotency_key being generated earlier - that way, table reader can re-use this path

lib/sequin/runtime/message.ex
```diff
@@ -39,10 +39,12 @@ defmodule Sequin.Runtime.SlotProcessor.Message do
     field :old_fields, list(Field.t())
     field :fields, list(Field.t())
     field :subscription_ids, list(String.t())
+    field :idempotency_key, String.t()
```
acco, line 45: in order to thread idempotency_key through, needed to add to Message

lib/sequin/runtime/slot_processor_server.ex
```diff
@@ -1263,6 +1263,7 @@ defmodule Sequin.Runtime.SlotProcessorServer do
       msg
       | commit_lsn: state.current_xaction_lsn,
         commit_idx: state.current_commit_idx,
+        idempotency_key: Base.encode64("#{state.current_xaction_lsn}:#{state.current_commit_idx}"),
```
acco: We set this higher up, on ingestion

lib/sequin/runtime/table_reader.ex
```diff
@@ -379,94 +379,34 @@ defmodule Sequin.Runtime.TableReader do
...
-    }
-  end
-
-  defp message_from_row(
-         %SinkConsumer{message_kind: :event} = consumer,
-         %Backfill{} = backfill,
-         %PostgresDatabaseTable{} = table,
-         record
-       ) do
+  defp message_from_row(%SinkConsumer{} = consumer, %Backfill{} = backfill, %PostgresDatabaseTable{} = table, record) do
```
acco, line 382: Now, we can rely on Consumers.consumer_message

lib/sequin/runtime/table_reader.ex
```diff
@@ -379,94 +379,34 @@ defmodule Sequin.Runtime.TableReader do
...
-      action: :read,
-      record: build_record_payload(table, record_attnums_to_values),
-      metadata: %ConsumerRecordData.Metadata{
-        database_name: consumer.replication_slot.postgres_database.name,
-        table_name: table.name,
+    message =
+      %Message{
+        action: :read,
+        errors: nil,
+        ids: record_pks,
```
acco, line 390: kind of a hack, but to use Consumers.consumer_message, we make a Message struct

We'll simplify later when we remove intermediate structs

lib/sequin/runtime/table_reader.ex
```diff
@@ -379,94 +379,34 @@ defmodule Sequin.Runtime.TableReader do
...
+            value = Map.fetch!(record, column.attnum)
+
+            %Message.Field{
+              column_name: column.name,
+              column_attnum: column.attnum,
+              value: value
+            }
+          end),
+        trace_id: UUID.uuid4(),
+        idempotency_key: Base.encode64(Enum.join([backfill.id | record_pks], ":"))
```
acco, line 406: idempotency key is generated differently here -

note i fixed an issue with how we generate (I think we want record_pks to be separted by : everywhere?)

PR #1966, opened 2025-07-09
✨ Load GCP credentials using application default credentials

davoclavo:


lib/sequin/gcp/application_default_credentials.ex
```diff
@@ -0,0 +1,273 @@
...
+  @doc """
+  Gets GCP credentials from the GOOGLE_APPLICATION_CREDENTIALS environment variable.
+
+  This function is kept for backward compatibility and testing purposes.
+  It's recommended to use `get_credentials/0` instead.
+
+  Returns `{:ok, credentials}` on success or `{:error, reason}` on failure.
+  """
+  @spec get_service_account_from_env() :: {:ok, map()} | {:error, term()}
+  def get_service_account_from_env do
```
davoclavo, line 68: Get rid of this and only rely on goth

test/sequin/gcp/application_default_credentials_test.exs
```diff
@@ -0,0 +1,177 @@
+defmodule Sequin.Gcp.ApplicationDefaultCredentialsTest do
```
davoclavo, line 1: Delete this test suite

PR #1973, opened 2025-07-10
auto assign on commenting take

ajita-asthana:
Simplify the process of assigning the issue to the contributor, user's can self assign the issue. 
An additional GitHub workflow that would run in response to work take submitted as a comment


acco:
Thanks @ajita-asthana – let's revisit later. We're still figuring out our workflow for contributions. After you work a couple issues, I'm sure you'll have great thoughts on how we can improve :)

PR #1999, opened 2025-07-15
🏎️ SQS performance review

davoclavo:


assets/svelte/consumers/SinkConsumerForm.svelte
```diff
@@ -424,6 +424,59 @@
             may improve.
           </p>
         </div>
+
+        <Accordion class="w-full">
```
thisisgoldman, line 428: I'd propose we move batching to the **Sink configuration** card similar to how we do it webhooks sinks. Batching is less about "grouping" (which is more to do with strict ordering) as it is about how we deliver payloads of messages to the destination.

Alternatively, we could move batching into it's own card?

CleanShot 2025-07-15 at 11 40 23@2x

acco:
@davoclavo What's the status of this one - still draft?

davoclavo:
@acco this is still in draft, I had some questions around what is the baseline performance for SQS as I was not able to get +10K msg/s using multiple different configurations. Happy to pick this up again tho.

PR #2001, opened 2025-07-15
✨ Direct SQL without substitutions

RTLS:


docs/reference/enrichment.mdx
```diff
@@ -0,0 +1,158 @@
+---
+title: SQL Enrichment
```
acco, line 2: What do we call this? 🤔

docs/reference/enrichment.mdx
```diff
@@ -0,0 +1,158 @@
...
+4. Send the enriched messages through the rest of your sink pipeline, including into transform and routing functions
+
+## SQL enrichment syntax
+
+SQL enrichment functions must:
+
+1. Use parameterized queries with `$1` syntax
+2. Select all primary key columns from the source table
+3. Select additional fields for enrichment
+4. Return 0 or 1 row per message
```
acco, line 28:
```suggestion
4. Return 0 rows or 1 row per message
```

docs/reference/enrichment.mdx
```diff
@@ -0,0 +1,158 @@
...
+  u.id, -- you must select all primary keys of the source table for Sequin to associate the enrichment with the message
+  a.name as account_name -- example of an enrichment field
+FROM
+  users u -- this enrichment is triggered on `users` table change messages
+JOIN
+  accounts a on u.account_id = a.id
+WHERE
+  -- the `ANY syntax is required for Sequin to perform batched queries
+  u.id = ANY($1)
+```
```
acco, line 43: This is great. Maybe even use a comment for $1:

```suggestion
```sql
SELECT
  u.id, -- you must select all primary keys of the source table for Sequin to associate the enrichment with the message
  a.name as account_name -- example of an enrichment field
FROM
  users u -- this enrichment is triggered on `users` table change messages
JOIN
  accounts a on u.account_id = a.id
WHERE
  -- the `ANY syntax is required for Sequin to perform batched queries
  u.id = ANY($1) -- $1 will be bound to a list of primary keys of users
```
```

File: docs/reference/enrichment.mdx [hunk 4]
```diff
@@ -0,0 +1,158 @@
...
+
+### Parameterization
+
+SQL enrichment functions must use the `$1` syntax for parameterization. This allows Sequin to safely batch multiple messages into a single query.
+
+#### Composite primary keys
+
+If your table has componsite primary keys, you must use a relatively advanced SQL where clause to select rows with those keys. Consider a table with a primary keys of `(id, account_id)`:
+
+```sql highlight={8-12}
```
acco, line 70: cool!

PR #2003, opened 2025-07-15
⏯️ Pause backfills + select query timeout

davoclavo:


assets/svelte/components/BackfillForm.svelte
```diff
@@ -211,6 +211,23 @@
       </p>
     </div>
 
+    <div class="space-y-2">
+      <Label for="maxTimeoutMs" class="text-sm">Query Timeout (ms)</Label>
```
acco:
```suggestion
      <Label for="maxTimeoutMs" class="text-sm">Query timeout (ms)</Label>
```

assets/svelte/components/BackfillForm.svelte
```diff
@@ -211,6 +211,23 @@
...
+        id="maxTimeoutMs"
+        bind:value={form.maxTimeoutMs}
+        min="1000"
+        max="600000"
+        step="1000"
+        class="w-48"
+        placeholder="5000"
+      />
+      <p class="text-xs text-muted-foreground">
+        Maximum time for each query during backfill (1-600 seconds)
```
acco:
```suggestion
        Maximum time for each <code>select</code> query performed during backfill (1-600 seconds)
```

assets/svelte/components/BackfillForm.svelte
```diff
@@ -211,6 +211,23 @@
...
+      </p>
+    </div>
```
acco: Probably need an info tooltip that links to the docs. The docs can say something about our PageSizeOptimizer: Sequin will automatically tune batch sizes in its select queries to maximize throughput during backfills (i.e. messages per second). This setting controls the maximum amount of time each `select` query should take. Sequin will decrease page sizes as necessary to keep `select` queries below this limit. If you're unsure, the default is fine to start.

lib/sequin/consumers/backfill.ex
```diff
@@ -74,11 +78,48 @@ defmodule Sequin.Consumers.Backfill do
...
+      {:active, :completed} -> true
+      # Can fail from active state
+      {:active, :failed} -> true
+      # Can cancel from active or paused state
+      {:active, :cancelled} -> true
+      {:paused, :cancelled} -> true
+      # All other transitions are invalid
+      _ -> false
+    end
   end
```
acco: 👍

lib/sequin/consumers/sink_consumer.ex
```diff
@@ -104,7 +104,7 @@ defmodule Sequin.Consumers.SinkConsumer do
 
     field :health, :map, virtual: true
 
-    has_many :active_backfills, Backfill, where: [state: :active]
+    has_many :active_backfills, Backfill, where: [state: {:in, [:active, :paused]}]
```
acco, line 107: 👍

lib/sequin/runtime/supervisor.ex
```diff
@@ -73,20 +73,50 @@ defmodule Sequin.Runtime.Supervisor do
...
+                Process.demonitor(ref, [:flush])
+                Process.exit(pid, :kill)
+                :ok
+            end
+
+          error ->
+            error
+        end
+    end
   end
```
acco: Ah were we terminating too aggressively?

lib/sequin/runtime/table_reader_server.ex
```diff
@@ -947,6 +973,29 @@ defmodule Sequin.Runtime.TableReaderServer do
...
+  rescue
+    error ->
+      Logger.error("[TableReaderServer] Error in terminate callback", error: inspect(error))
+      :ok
+  end
+
+  def terminate(_reason, _state_name, _state) do
+    # Handle case where state is not yet initialized
+    :ok
+  end
```
acco: Cool - as we discussed, not necessary for v1

We don't trap exits for graceful shutdown anywhere else, but that doesn't mean we shouldn't. I'm curious what pattern we should adopt here.

lib/sequin/runtime/table_reader_server.ex
```diff
@@ -320,21 +324,37 @@ defmodule Sequin.Runtime.TableReaderServer do
...
+      check_sms_timeout(state.setting_check_sms_timeout)
     ]
 
+    # Only schedule initial fetch if backfill is active
+    actions =
+      if backfill.state == :active do
+        actions ++ [maybe_fetch_timeout()]
+      else
+        actions
+      end
```
acco: I think it's safe to assume if we're here that we're :active

If not, then I'd consider raising way up at the top when we load the backfill (if you want to be strict about races)

lib/sequin/runtime/table_reader_server.ex
```diff
@@ -320,21 +324,37 @@ defmodule Sequin.Runtime.TableReaderServer do
         backfill.initial_min_cursor
       end
 
+    # Update page_size_optimizer with backfill's max_timeout_ms
+    page_size_optimizer =
+      state.page_size_optimizer_mod.new(
+        initial_page_size: state.page_size_optimizer_mod.size(state.page_size_optimizer),
+        max_timeout_ms: backfill.max_timeout_ms || to_timeout(second: 5),
+        max_page_size: 40_000
+      )
```
acco: We should remove the initialization of page_size_optimizer earlier on then, in `init/1`?

https://github.com/sequinstream/sequin/pull/2003/files#diff-779eaaa26838c6e369104a0dfdd4094d7ab9a42ad424a32e4c1a317abc4da664R262-R267

lib/sequin_web/live/sink_consumers/show.ex
```diff
@@ -401,6 +403,52 @@ defmodule SequinWeb.SinkConsumersLive.Show do
     end
   end
 
+  def handle_event("pause_backfill", %{"backfill_id" => backfill_id}, socket) do
+    case Consumers.get_backfill(backfill_id) do
+      {:ok, backfill} ->
+        if backfill.sink_consumer_id == socket.assigns.consumer.id do
```
acco: This is good protection. Other thing you can do is just blow up if they don't match (this means either a (a) bug on our service or (b) malicious user

```suggestion
    # bind consumer_id earlier
    case Consumers.get_backfill(backfill_id) do
      {:ok, %{ consumer_id: ^consumer_id } ->
        if backfill.sink_consumer_id == socket.assigns.consumer.id do
```

Or we can add a new function `get_backfill_for_consumer(consumer_id, backfill_id)`

The reason I like these two approaches is that it's way clearer to the reader that the backfill_id _not_ belonging to the consumer is exceptional

acco: And then with a `with` clause this can become one happy path + one `else` fall-through

lib/sequin_web/live/sink_consumers/show.ex
```diff
@@ -401,6 +403,52 @@ defmodule SequinWeb.SinkConsumersLive.Show do
...
+      {:error, _error} ->
+        {:reply, %{ok: false}, put_flash(socket, :toast, %{kind: :error, title: "Backfill not found"})}
+    end
+  end
+
+  def handle_event("resume_backfill", %{"backfill_id" => backfill_id}, socket) do
+    case Consumers.get_backfill(backfill_id) do
+      {:ok, backfill} ->
+        if backfill.sink_consumer_id == socket.assigns.consumer.id do
+          case Consumers.update_backfill(backfill, %{state: :active}) do
```
acco: ditto as above

priv/repo/migrations/20250715000002_add_paused_state_to_backfills.exs
```diff
@@ -0,0 +1,52 @@
...
+
+  def up do
+    # Add paused state to backfill state enum
+    execute "ALTER TYPE #{@config_schema}.backfill_state ADD VALUE 'paused';"
+  end
+
+  def down do
+    # Attempt to remove the 'paused' value from the enum
+    # This will fail if there are any backfills in the 'paused' state
+    execute """
```
acco: Oh man, hopefully this was zero cost to write!

We just simply don't drop values from types in down migrations, not worth the pain

priv/repo/migrations/20250715000002_add_paused_state_to_backfills.exs
```diff
@@ -0,0 +1,52 @@
+defmodule Sequin.Repo.Migrations.AddPausedStateToBackfills do
+  use Ecto.Migration
+
+  @config_schema Application.compile_env(:sequin, [Sequin.Repo, :config_schema_prefix])
+
+  def up do
+    # Add paused state to backfill state enum
+    execute "ALTER TYPE #{@config_schema}.backfill_state ADD VALUE 'paused';"
+  end
```
acco: Per below, because we don't drop, you can also make this idempotent

priv/repo/migrations/20250715000003_update_backfills_index_for_paused_state.exs
acco, line 1: Hmm I don't get this change. We still want to allow for only one :active backfill at a time yeah?

test/sequin_web/live/sink_consumers_live_show_test.exs
```diff
@@ -125,5 +125,112 @@ defmodule SequinWeb.SinkConsumersLive.ShowTest do
 
       assert backfill.initial_min_cursor == %{1 => "1"}
     end
+
+    test "runs a backfill with custom max timeout", %{conn: conn, consumer: consumer, table: table} do
```
acco, line 129: thorough!

test/sequin/consumers/backfill_test.exs
```diff
@@ -108,5 +108,145 @@ defmodule Sequin.Consumers.BackfillTest do
 
       assert %{sink_consumer_id: ["already has an active backfill"]} = errors_on(changeset)
     end
+
+    test "allows transitioning backfill to paused state" do
```
acco: I'm not sure we need all these tests! They're just a hair above "Ecto works" -- it's just reimplementing the same business logic in a second place

I think your integration tests for liveview cover the 80/20 nicely

Which ones do you think are essential?

test/sequin/consumers_test.exs
```diff
@@ -806,4 +806,130 @@ defmodule Sequin.ConsumersTest do
       assert group_id_from_message == "user_1:group_A"
     end
   end
+
+  describe "update_backfill/2 for pausing and resuming" do
```
acco: Ditto here - may not need all of these

test/sequin/table_reader_server_test.exs
```diff
@@ -508,6 +508,175 @@ defmodule Sequin.Runtime.TableReaderServerTest do
       assert record["power_level"] == 9001
     end
 
+    @tag :capture_log
+    test "handles paused state correctly", %{backfill: backfill, table_oid: table_oid, consumer: consumer} do
```
acco, line 513: Way to figure these out! Not for faint of heart

test/sequin/table_reader_server_test.exs
```diff
@@ -508,6 +508,175 @@ defmodule Sequin.Runtime.TableReaderServerTest do
...
+          page_size_optimizer_mod: nil,
+          check_state_timeout: 50,
+          initial_page_size: 2
+        )
+
+      # Verify the process is alive
+      assert Process.alive?(table_reader_pid)
+
+      # Give the process a moment to start and check if it should fetch
+      Process.sleep(100)
```
acco: Not necessary given assert_receive that follows

test/sequin/table_reader_server_test.exs
```diff
@@ -508,6 +508,175 @@ defmodule Sequin.Runtime.TableReaderServerTest do
...
+      # Start the table reader with short check_state_timeout for testing and small page size
+      table_reader_pid =
+        start_table_reader_server(backfill, table_oid,
+          page_size_optimizer_mod: nil,
+          check_state_timeout: 50,
+          initial_page_size: 2
+        )
+
+      # Verify the process is alive
+      assert Process.alive?(table_reader_pid)
```
acco: not necessary

test/sequin/table_reader_server_test.exs
```diff
@@ -508,6 +508,175 @@ defmodule Sequin.Runtime.TableReaderServerTest do
...
+              100 -> 1
+            end
+        after
+          0 -> 0
+        end
+
+      # If we cleared messages, wait a bit more to ensure no new fetches are scheduled
+      if cleared > 0 do
+        Process.sleep(300)
+      end
```
acco: I think this can all become:

```
assert_eventually !Process.alive?(table_reader_pid)
```

or something

test/sequin/table_reader_server_test.exs
```diff
@@ -508,6 +508,175 @@ defmodule Sequin.Runtime.TableReaderServerTest do
...
+
+      # Wait and verify no new batches are fetched while paused
+      refute_receive {TableReaderServer, {:batch_fetched, _batch_id}}, 2000
```
acco: Probably don't need with adjustment above

Note this will block for 2s

test/sequin/table_reader_server_test.exs
```diff
@@ -508,6 +508,175 @@ defmodule Sequin.Runtime.TableReaderServerTest do
...
+      _table_reader_pid =
+        start_table_reader_server(
+          updated_backfill,
+          table_oid,
+          page_size_optimizer_mod: PageSizeOptimizerMock
+        )
+
+      # Wait for init to complete and fetch to start
+      assert_receive {TableReaderServer, {:batch_fetched, _batch_id}}, 5000
+    end
```
acco: You might be able to just assert this in a prior test around PageSizeOptimizer?

test/sequin/table_reader_server_test.exs
```diff
@@ -508,6 +508,175 @@ defmodule Sequin.Runtime.TableReaderServerTest do
...
+
+      # Let it fetch some initial messages
+      assert_receive {TableReaderServer, {:batch_fetched, _batch_id}}, 5000
+
+      # Pause the backfill
+      {:ok, _} = Consumers.update_backfill(backfill, %{state: :paused})
+
+      # Process should stop when it detects the paused state
+      assert_receive {:DOWN, ^ref, :process, ^table_reader_pid, :normal}, 1000
+    end
```
acco: I think this is captured by prior test just fine

acco (changes requested):

lib/sequin/consumers/consumers.ex
```diff
@@ -1709,7 +1709,10 @@ defmodule Sequin.Consumers do
     |> Repo.one()
   end
 
-  def create_backfills_for_form(account_id, consumer, tables_config) do
+  def create_backfills_for_form(account_id, consumer, tables_config, max_timeout_ms \\ 5000) do
+    # Ensure postgres_database association is loaded
+    consumer = Repo.preload(consumer, :postgres_database)
```
davoclavo: do we really need this?

lib/sequin/runtime/supervisor.ex
```diff
@@ -73,20 +73,50 @@ defmodule Sequin.Runtime.Supervisor do
...
-    :ok
+  def stop_table_reader(supervisor, %Backfill{id: backfill_id}, opts) do
+    stop_table_reader(supervisor, backfill_id, opts)
   end
 
-  def stop_table_reader(supervisor, backfill_id) when is_binary(backfill_id) do
-    Sequin.DynamicSupervisor.stop_child(supervisor, TableReaderServer.via_tuple(backfill_id))
-    :ok
+  def stop_table_reader(supervisor, backfill_id, opts) when is_binary(backfill_id) do
+    # Default timeout of 5 seconds for graceful shutdown
```
davoclavo: get rid of these changes regarding graceful shutdown

lib/sequin/runtime/supervisor.ex
davoclavo: @acco There is an issue with the metrics being reported, after some research this is what I thought could be the culprit, but I am still unsure. We can re-visit this later

assets/svelte/components/BackfillForm.svelte
davoclavo: 👍🏼 Added documentation and a link to the docs. Sadly couldn't add the on-hover Tooltip as being inside a Modal behaves weird 😢  - reminds me of the Full Page Modal issues

test/sequin/consumers_test.exs
davoclavo: Agreed. Some of these were useful to validate allowed status transitions (e.g. completed -> paused should fail), will trim them to the minimum tho.

priv/repo/migrations/20250715000002_add_paused_state_to_backfills.exs
davoclavo: So just to confirm I understand, you are suggesting we just perform a no-op in the down fn?

priv/repo/migrations/20250715000003_update_backfills_index_for_paused_state.exs
davoclavo: I initially thought `paused` + `active` both fall under the "active" category. Perhaps I am wrong, should we allow a single `active` but potentially multiple `paused`? or a single `active`|`paused` backfill?

test/sequin/consumers/backfill_test.exs
davoclavo: I decided to keep:

1. State transition test
2. Unique index test

assets/svelte/components/BackfillForm.svelte
acco: Ah yes!

priv/repo/migrations/20250715000003_update_backfills_index_for_paused_state.exs
acco: Ah, I see! Hmm...

Two options, both are equal IMO:

1.

- only one active allowed
- many paused allowed

This means we need some validation around un-pausing.

2.

- only one active _or_paused
- If you try to start a new backfill when one is active/paused, the button is disabled or whatever

docs/reference/backfills.mdx
```diff
@@ -63,6 +63,14 @@ By default, backfills order rows by primary keys of the table. This ensures that
 
 For partial backfills, you choose a sort column. Sequin will order rows by the sort column. You may choose any column as the sort column, but we strongly recommend using a column that is indexed.
 
+## Query timeout
+
+Sequin automatically tunes batch sizes in its select queries to maximize throughput during backfills (i.e. messages per second). This setting controls the maximum amount of time each select query should take.
+
+Sequin uses a PageSizeOptimizer that will decrease page sizes as necessary to keep select queries below this limit. This ensures consistent performance and prevents timeouts even when processing large batches of data.
+
+If you're unsure, the default of 5000ms (5 seconds) is fine to start.
```
acco, line 72:
```suggestion
## Query timeout

Sequin automatically tunes batch sizes in its `select` queries to maximize throughput during backfills (i.e. maximize messages per second). It does so while respecting a max query timeout limit of 5 seconds.

You can tune this timeout limit when starting a backfill. Setting to a lower value may decrease database contention but impact throughput.
```

lib/sequin_web/live/sink_consumers/show.ex
```diff
@@ -1650,6 +1624,29 @@ defmodule SequinWeb.SinkConsumersLive.Show do
...
+       |> load_consumer_backfills()}
+    else
+      {:error, %NotFoundError{}} ->
+        {:reply, %{ok: false}, put_flash(socket, :toast, %{kind: :error, title: "Backfill not found"})}
+
+      {:error, _error} ->
+        {:reply, %{ok: false},
+         put_flash(socket, :toast, %{kind: :error, title: "Failed to #{format_state_action(target_state)} backfill"})}
+    end
+  end
```
acco: awesome!

test/sequin/consumers/backfill_test.exs
```diff
@@ -108,5 +108,55 @@ defmodule Sequin.Consumers.BackfillTest do
...
+      assert {:ok, cancelled_backfill} =
+               Repo.update(Backfill.update_changeset(resumed_backfill, %{state: :cancelled}))
+
+      assert cancelled_backfill.state == :cancelled
+
+      assert {:error, _changeset} =
+               Repo.update(Backfill.update_changeset(cancelled_backfill, %{state: :active}))
+    end
+
+    test "prevents multiple active or paused backfills for the same consumer and table" do
```
acco, line 135: These are great, nice whittling down :+1

acco (approved):

priv/repo/migrations/20250715000003_update_backfills_index_for_paused_state.exs
davoclavo: I think a single `active | pending` is a good path forward, as it would be odd to backfill multiple times the same table, unless they were for mutually exclusive sections of a table - but that's a bit trickier to solve?

And regarding your suggestions, I would lean towards Option #2, for the reason stated above - however there is a case which will be tricky to solve:

For example: there is already a backfill for Table A (active or paused), and the user wants to perform a backfill for Table B, and both tables are tracked by the sink. It should still go through, and the button should not be disabled once Table A is paused|active.

So I suggest just showing proper error messages in the toast notification stating why an action failed, instead of disabling the button. What do you think?

PR #2009, opened 2025-07-16
🥏 Discard failing | all pending messages in message tab

### LINKED ISSUE #1435: Add “Discard All” button to Messages tab for bulk-discarding messages
### **Context / Background**

The **Messages** tab in the Sequin console currently lets users:

* View **in-progress**, **delivering**, and **recently-acknowledged** messages.
* Trigger **Re-deliver All**, which immediately retries every pending message (useful when exponential-back-off has kicked in).

During development (and occasionally in production) users need a quick way to **purge stuck or failing messages** to give their sync a clean slate.

---

### **Problem**

There is **no UI control** to discard multiple messages at once. Users must:

1. Manually mark individual messages as discarded (click message, then click "Acknowledge" button in the slide-out), *or*
2. Set `max_retry_count` and wait until messages exceed it.

Both options are slow.

---

### **Solution / Scope**

Introduce a **“Discard All”** control to the "Messages" tab with confirmation dialog that can:

* **Discard failing messages only** (messages currently in a failed state).
* **Discard all messages** (regardless of state).

---

### **User Experience**

1. **Button placement**
   * Next to the existing **Re-deliver All** button in the Messages tab action bar.
2. **Confirmation dialog**
   * **Title:** “Discard messages”
   * **Radio buttons (mutually exclusive):**
     * “Discard *failing* messages” (default)
     * “Discard *all* messages”
   * **Destructive action button**
     * Label mirrors selection: **“Discard Failing”** or **“Discard All”**
     * Styled as a prominent red/destructive button
   * Secondary **Cancel** button closes dialog without action.
3. **Feedback**
   * On success, flash a green toast:
     * “Discarded N failing messages” — *or* — “Discarded N messages.”
       * (No need for "N" if not easy to get)
   * On error, show red toast with error text.

---

### **Backend Requirements**

* **SlotMessageStore** (`slot_message_store.ex`):
  * Add two new public functions & callback handlers:
    * `discard_all_messages/1`
    * `discard_failing_messages/1`
  * Re-use existing *discard* logic already exercised in SlotMessageStore when `ack_failed` exceeds `max_retry_count`.

---

### **Implementation Plan**

| Stage | Area | Tasks |
| -- | -- | -- |
| **1** | **Backend** | • Implement the two discard functions in **SlotMessageStore**.• Unit-test discarding scenarios (no messages, some messages, concurrency). |
| **2** | **Frontend** | • Add “Discard All” button + confirmation dialog.• Wire dialog choices to new backend endpoints.• Re-use pattern used by **Re-deliver All** for optimistic UI and toasts. |

---

### **Acceptance Criteria**

1. **UI**
   * Button and confirmation dialog render in Messages tab.
2. **Discard failing**
   * Selecting “Discard failing messages” removes only messages in failed state.
3. **Discard all**
   * Selecting “Discard all messages” removes *every* message visible in table.
4. **Safety**
   * Destructive action requires confirmation; clicking **Cancel** does nothing.
5. **Feedback**
   * Success and error toasts appear (with correct counts / messages, if applicable).

---

### **Out of Scope (for this ticket)**

* YAML / RPC API support for discarding.
* Per-message discard operations (already available).

---

### **Notes & References**

* **Discard logic already exists** in `ack_failed` path (see `max_retry_count` handling).
* Follow the same **coding conventions** and **test helpers** used by the Re-deliver All implementation.

---

davoclavo:
Add button to discard all failing or pending messages.

image

Note:
The current implementation discards all currently *loaded/buffered* messages, but there might be some messages that did not yet fit in our internal buffer, thus aren't discarded. So in case a Sink is processing thousands of messages and you want to discard them **all**, you'd need to discard buffered portions multiple times.

Do we want a different mechanism? or is that OK?

Update: I guess that is OK, if the user wants to discard any upcoming messages they could disable the sink.

---

Fixes #1435

### DIFF
--- a/lib/sequin/runtime/slot_message_store.ex
+++ b/lib/sequin/runtime/slot_message_store.ex
@@ -765,13 +765,29 @@ defmodule Sequin.Runtime.SlotMessageStore do
   @decorate track_metrics("discard_all_messages")
   def handle_call(:discard_all_messages, _from, state) do
     all_messages = State.all_messages(state)
-    do_discard_messages(state, all_messages)
+    all_ack_ids = Enum.map(all_messages, & &1.ack_id)
+
+    case handle_call({:messages_succeeded, all_ack_ids, false}, nil, state) do
+      {:reply, {:ok, count, _ack_ids}, new_state} ->
+        {:reply, {:ok, count}, new_state}
+
+      other ->
+        other
+    end
   end
 
   @decorate track_metrics("discard_failing_messages")
   def handle_call(:discard_failing_messages, _from, state) do
     failing_messages = State.failing_messages(state)
-    do_discard_messages(state, failing_messages)
+    failing_ack_ids = Enum.map(failing_messages, & &1.ack_id)
+
+    case handle_call({:messages_succeeded, failing_ack_ids, false}, nil, state) do
+      {:reply, {:ok, count, _ack_ids}, new_state} ->
+        {:reply, {:ok, count}, new_state}
+
+      other ->
+        other
+    end
   end
 
   @decorate track_metrics("min_unpersisted_wal_cursor")
@@ -1135,25 +1151,6 @@ defmodule Sequin.Runtime.SlotMessageStore do
   defp handle_discarded_messages(%State{} = state, discarded_messages) do
     with :ok <- delete_messages(state, Enum.map(discarded_messages, & &1.ack_id)) do
       AcknowledgedMessages.store_messages(state.consumer.id, discarded_messages)
-    end
-  end
-
-  # Common logic for discarding messages through the existing ack path
-  defp do_discard_messages(state, messages) do
-    {all_ack_ids, discarded_messages} = State.prepare_messages_for_discard(messages)
-
-    case AcknowledgedMessages.store_messages(state.consumer.id, discarded_messages) do
-      :ok ->
-        case handle_call({:messages_succeeded, all_ack_ids, false}, nil, state) do
-          {:reply, {:ok, count, _ack_ids}, new_state} ->
-            {:reply, {:ok, count}, new_state}
-
-          other ->
-            other
-        end
-
-      error ->
-        {:reply, error, state}
     end
   end
 
--- a/lib/sequin/runtime/slot_message_store_state.ex
+++ b/lib/sequin/runtime/slot_message_store_state.ex
@@ -554,32 +554,12 @@ defmodule Sequin.Runtime.SlotMessageStore.State do
   end
 
   @doc """
-  Returns messages that are currently failing (have not_visible_until set in the future).
+  Returns messages that are currently failing (have been delivered at least once).
   """
   @spec failing_messages(State.t()) :: list(message())
   def failing_messages(%State{} = state) do
-    now = Sequin.utc_now()
-
     state.messages
     |> Map.values()
-    |> Enum.filter(fn msg ->
-      not is_nil(msg.not_visible_until) and DateTime.after?(msg.not_visible_until, now)
-    end)
-  end
-
-  @doc """
-  Prepares messages for discard by setting their state to :discarded and clearing not_visible_until.
-  Returns a tuple of {all_ack_ids, discarded_messages} where:
-  - all_ack_ids: ack_ids of all messages to remove from memory
-  - discarded_messages: all messages with their state set to :discarded
-  """
-  @spec prepare_messages_for_discard(list(message())) ::
-          {list(SinkConsumer.ack_id()), list(message())}
-  def prepare_messages_for_discard(messages) do
-    messages
-    |> Enum.map(fn msg ->
-      {msg.ack_id, %{msg | state: :discarded, not_visible_until: nil}}
-    end)
-    |> Enum.unzip()
+    |> Enum.filter(fn msg -> msg.deliver_count > 0 end)
   end
 end
--- a/test/sequin/slot_message_store_test.exs
+++ b/test/sequin/slot_message_store_test.exs
@@ -1071,7 +1071,8 @@ defmodule Sequin.SlotMessageStoreTest do
           ConsumersFactory.consumer_message(
             message_kind: consumer.message_kind,
             consumer_id: consumer.id,
-            group_id: "test-group-#{i}"
+            group_id: "test-group-#{i}",
+            deliver_count: 0
           )
         end)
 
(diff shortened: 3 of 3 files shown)


lib/sequin_web/live/sink_consumers/show.ex
```diff
@@ -590,6 +590,46 @@ defmodule SequinWeb.SinkConsumersLive.Show do
...
 
+  def handle_event("discard_messages", %{"discard_type" => discard_type}, socket) do
+    consumer = socket.assigns.consumer
+
+    {result, success_message} =
+      case discard_type do
+        "failing" ->
+          case SlotMessageStore.discard_failing_messages(consumer) do
+            {:ok, count} ->
+              {{:ok, count}, "Discarded #{count} failing message#{if count == 1, do: "", else: "s"}"}
```
davoclavo: `failing message(s)`

lib/sequin/runtime/slot_message_store.ex
```diff
@@ -990,6 +1045,21 @@ defmodule Sequin.Runtime.SlotMessageStore do
...
+      case GenServer.call(via_tuple(consumer.id, partition), call_name) do
+        {:ok, count} -> {:cont, {:ok, acc_count + count}}
+        error -> {:halt, error}
+      end
+    end)
+  catch
+    :exit, e ->
+      {:error, exit_to_sequin_error(e)}
+  end
+
```
acco, line 1051: Great

lib/sequin/runtime/slot_message_store.ex
```diff
@@ -746,6 +762,45 @@ defmodule Sequin.Runtime.SlotMessageStore do
...
+    case discard_messages_and_update_state(state, failing_messages) do
+      {:ok, count, new_state} ->
+        Sequin.ProcessMetrics.gauge("message_count", map_size(new_state.messages))
+        Sequin.ProcessMetrics.gauge("payload_size_bytes", new_state.payload_size_bytes)
+        {:reply, {:ok, count}, new_state}
+
+      error ->
+        {:reply, error, state}
+    end
+  end
```
acco: - Move as much as possible down to `State`
- We want to route through the same path as how we ack messages. I'm not sure how we do that. Can we simply get a list of IDs for messages we want to discard and then reuse an existing path?

lib/sequin/runtime/slot_message_store.ex
RTLS: yep great improvement

lib/sequin/runtime/slot_message_store.ex
davoclavo: @acco good idea

Do you suggest to send these messages through the succesfull messages path?

https://github.com/sequinstream/sequin/blob/davoclavo/feat/add_discard_all_button/lib/sequin/runtime/slot_message_store.ex#L643-L669


or failed messages path?

https://github.com/sequinstream/sequin/blob/davoclavo/feat/add_discard_all_button/lib/sequin/runtime/slot_message_store.ex#L703-L739

---

I guess `messages_succeeded` to avoid potential failure handlers that may retry a message, but not sure if I may be missing something

acco: @davoclavo `messages_succeeded` path!

lib/sequin/runtime/slot_message_store.ex
```diff
@@ -1095,6 +1138,25 @@ defmodule Sequin.Runtime.SlotMessageStore do
     end
   end
 
+  # Common logic for discarding messages through the existing ack path
+  defp do_discard_messages(state, messages) do
+    {all_ack_ids, discarded_messages} = State.prepare_messages_for_discard(messages)
+
+    case AcknowledgedMessages.store_messages(state.consumer.id, discarded_messages) do
```
acco: I think we can safely skip AcknowledgedMessages storage - but great catch!

lib/sequin/runtime/slot_message_store.ex
```diff
@@ -1095,6 +1138,25 @@ defmodule Sequin.Runtime.SlotMessageStore do
...
+      :ok ->
+        case handle_call({:messages_succeeded, all_ack_ids, false}, nil, state) do
+          {:reply, {:ok, count, _ack_ids}, new_state} ->
+            {:reply, {:ok, count}, new_state}
```
acco: 👍

lib/sequin/runtime/slot_message_store_state.ex
```diff
@@ -544,4 +544,42 @@ defmodule Sequin.Runtime.SlotMessageStore.State do
...
+  @spec failing_messages(State.t()) :: list(message())
+  def failing_messages(%State{} = state) do
+    now = Sequin.utc_now()
+
+    state.messages
+    |> Map.values()
+    |> Enum.filter(fn msg ->
+      not is_nil(msg.not_visible_until) and DateTime.after?(msg.not_visible_until, now)
+    end)
+  end
```
acco: We can perhaps just do `deliver_count` > 0?

lib/sequin/runtime/slot_message_store_state.ex
```diff
@@ -544,4 +544,42 @@ defmodule Sequin.Runtime.SlotMessageStore.State do
...
+  """
+  @spec prepare_messages_for_discard(list(message())) ::
+          {list(SinkConsumer.ack_id()), list(message())}
+  def prepare_messages_for_discard(messages) do
+    messages
+    |> Enum.map(fn msg ->
+      {msg.ack_id, %{msg | state: :discarded, not_visible_until: nil}}
+    end)
+    |> Enum.unzip()
+  end
```
acco: We might not need this anymore, if we're not putting in acknowledgedmessage store

So I think you can just do:

```
 def handle_call(:discard_all_messages, _from, state) do
  all_ack_ids = ...
  handle_call({:messages_succeeded, all_ack_ids, false}, nil, state)
end
```

lib/sequin/runtime/slot_message_store_state.ex
davoclavo: yep thats way better

lib/sequin/runtime/slot_message_store_state.ex
davoclavo: I took your suggestion and slightly changed the implementation of discarding to:

```elixir
  def handle_call(:discard_all_messages, _from, state) do
    all_messages = State.all_messages(state)
    all_ack_ids = Enum.map(all_messages, & &1.ack_id)

    case handle_call({:messages_succeeded, all_ack_ids, false}, nil, state) do
      {:reply, {:ok, count, _ack_ids}, new_state} ->
        {:reply, {:ok, count}, new_state}

      other ->
        other
    end
  end
```

as I don't currently see any value of returning the `ack_ids` as we only display the total count in the flash notification

acco (approved):

PR #2019, opened 2025-07-17
🚏 Add support for meilisearch functions

### LINKED ISSUE #2015: Support functions in Meilisearch
Meilisearch's functions capability allows for powerful ways to update collections. As an example, imagine we have a document:

```bash
curl -X PUT 'https://ms-732eba82ae36-26491.sfo.meilisearch.io/indexes/documents/documents' \
>      -H 'Content-Type: application/json' \
>      -H 'Authorization: Bearer secret' \
>      --data-binary '[
>        {
>          "id": 5,
>          "created": "2025-07-10T10:32:06Z",
>          "updated": "2025-07-10T10:32:06Z",
>          "content": [
>            "block1 - v1",
>            "block2 - v1",
>            "block3 - v1"
>          ]
>        }
>      ]'
```

We can update just the block at index `1` with a function:

```bash

curl -X POST 'https://ms-732eba82ae36-26491.sfo.meilisearch.io/indexes/documents/documents/edit' \
>      -H 'Content-Type: application/json' \
>      -H 'Authorization: Bearer secret' \
>      --data-binary '{
>        "filter": "id = 5",
>        "context": { "new_block": "block2 - v2" },
>        "function": "doc.content[1] = context.new_block"
>      }'
```

davoclavo:
Fixes #2015

### DIFF
--- a/assets/svelte/sinks/meilisearch/MeilisearchSinkForm.svelte
+++ b/assets/svelte/sinks/meilisearch/MeilisearchSinkForm.svelte
@@ -64,6 +64,26 @@
             </li>
           </ul>
         </div>
+        <div class="mb-2">
+          <strong>Advanced:</strong> You can use routing functions to perform
+          partial updates using Meilisearch's function capability:
+          <ul class="ml-6 list-disc">
+            <li>
+              Return <code>action: :function</code> from your routing function
+            </li>
+            <li>
+              Specify a filter expression and a <a
+                class="underline font-medium"
+                href="https://rhai.rs/book/engine/hello-world.html"
+                target="_blank">Rhai</a
+              > (a JavaScript-like language) function to update specific fields.
+            </li>
+            <li>
+              This is useful for updating nested arrays or specific document
+              properties
+            </li>
+          </ul>
+        </div>
       </Alert.Description>
     </Alert.Root>
 
--- a/lib/sequin/runtime/meilisearch_pipeline.ex
+++ b/lib/sequin/runtime/meilisearch_pipeline.ex
@@ -23,42 +23,39 @@ defmodule Sequin.Runtime.MeilisearchPipeline do
         concurrency: concurrency,
         batch_size: consumer.sink.batch_size || 100,
         batch_timeout: 5
-      ],
-      delete: [
-        concurrency: concurrency,
-        batch_size: consumer.sink.batch_size || 100,
-        batch_timeout: 5
       ]
     ]
   end
 
   @impl SinkPipeline
   def handle_message(message, context) do
-    %Routing.Consumers.Meilisearch{action: action, index_name: index_name} =
-      Routing.route_message(context.consumer, message.data)
+    routing_info = Routing.route_message(context.consumer, message.data)
+    %Routing.Consumers.Meilisearch{action: action, index_name: index_name} = routing_info
 
-    batcher =
-      case action do
-        :index -> :default
-        :delete -> :delete
+    action =
+      if is_atom(action) do
+        # Backwards compatibility with existing routing functions with atom actions
+        to_string(action)
+      else
+        action
       end
 
     message =
       message
-      |> Broadway.Message.put_batcher(batcher)
-      |> Broadway.Message.put_batch_key(index_name)
+      |> Broadway.Message.put_batch_key({action, index_name})
+      |> Broadway.Message.update_data(fn data ->
+        Map.put(data, :routing_info, routing_info)
+      end)
 
     {:ok, message, context}
   end
 
   @impl SinkPipeline
-  def handle_batch(:default, messages, batch_info, context) do
+  def handle_batch(:default, messages, %{batch_key: {"index", index_name}}, context) do
     %{
       consumer: %SinkConsumer{sink: sink, transform: transform} = consumer,
       test_pid: test_pid
     } = context
-
-    index_name = batch_info.batch_key
 
     setup_allowances(test_pid)
 
@@ -70,7 +67,7 @@ defmodule Sequin.Runtime.MeilisearchPipeline do
       end)
 
     case Client.import_documents(sink, index_name, records) do
-      {:ok} ->
+      :ok ->
         Trace.info(consumer.id, %Trace.Event{
           message: "Imported documents to \"#{index_name}\" index"
         })
@@ -88,13 +85,11 @@ defmodule Sequin.Runtime.MeilisearchPipeline do
   end
 
   @impl SinkPipeline
-  def handle_batch(:delete, messages, batch_info, context) do
+  def handle_batch(:default, messages, %{batch_key: {"delete", index_name}}, context) do
     %{
       consumer: %SinkConsumer{sink: sink} = consumer,
       test_pid: test_pid
     } = context
-
-    index_name = batch_info.batch_key
 
     setup_allowances(test_pid)
 
@@ -102,7 +97,7 @@ defmodule Sequin.Runtime.MeilisearchPipeline do
       Enum.flat_map(messages, fn %{data: message} -> message.record_pks end)
 
     case Client.delete_documents(sink, index_name, document_ids) do
-      {:ok} ->
+      :ok ->
         Trace.info(consumer.id, %Trace.Event{
           message: "Deleted documents from \"#{index_name}\" index",
           extra: %{document_ids: document_ids}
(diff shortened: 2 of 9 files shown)


assets/svelte/sinks/meilisearch/MeilisearchSinkForm.svelte
```diff
@@ -64,6 +64,23 @@
...
+        <div class="mb-2">
+          <strong>Advanced:</strong> You can use routing functions to perform
+          partial updates using Meilisearch's function capability:
+          <ul class="ml-6 list-disc">
+            <li>
+              Return <code>action: :function</code> from your routing function
+            </li>
+            <li>
+              Specify a filter expression and JavaScript-like function to update
+              specific fields
```
acco: This is Rhai right?

acco: Ah I see - it's specifically designed to be similar to JS. We can say something like "Rhai (a JavaScript-like language)" and link to their docs

lib/sequin/runtime/meilisearch_pipeline.ex
```diff
@@ -28,25 +28,34 @@ defmodule Sequin.Runtime.MeilisearchPipeline do
         concurrency: concurrency,
         batch_size: consumer.sink.batch_size || 100,
         batch_timeout: 5
+      ],
+      function: [
+        concurrency: concurrency,
+        batch_size: consumer.sink.batch_size || 100,
+        batch_timeout: 5
       ]
```
acco: Totally fine, but I think we use different batchers when we want to have different runtimes between them? Like concurrency, batch_size, batch_timeout

Otherwise I think we could just use pattern matching based on the batch_key? eg

```
batcher =
      case action do
        :index -> :default
        :delete -> :delete
        :function -> :function
      end

    message =
      message
      |> Broadway.Message.put_batch_key({batcher, index_name})
      |> Broadway.Message.update_data(fn data ->
        Map.put(data, :routing_info, routing_info)
      end)

# later on...

```elixir
def handle_batch(_batcher, messages, %{batch_key: {:function, index_name}}, context) do
end
```

i.e. they can all share the same "batcher" cluster

lib/sequin/runtime/meilisearch_pipeline.ex
```diff
@@ -121,6 +130,52 @@ defmodule Sequin.Runtime.MeilisearchPipeline do
...
+
+    index_name = batch_info.batch_key
+
+    setup_allowances(test_pid)
+
+    # Group messages by filter and function expression to batch similar updates
+    grouped_updates =
+      Enum.group_by(messages, fn %{data: %{routing_info: routing_info}} ->
+        {routing_info.filter, routing_info.function_expression, routing_info.context}
+      end)
```
acco: could potentially put all this in the batch_key 🤔  then we'd only be dealing with one batch at a time here.

lib/sequin/sinks/meilisearch/client.ex
```diff
@@ -108,6 +108,42 @@ defmodule Sequin.Sinks.Meilisearch.Client do
...
+    req =
+      sink
+      |> base_request()
+      |> Req.merge(
+        url: "/indexes/#{index_name}/documents/edit",
+        body: Jason.encode!(body),
+        headers: [{"Content-Type", "application/json"}]
+      )
+
+    case Req.post(req) do
```
acco, line 161: What kind of error do we get on non-200?

lib/sequin/sinks/meilisearch/client.ex
```diff
@@ -108,6 +108,42 @@ defmodule Sequin.Sinks.Meilisearch.Client do
     end
   end
 
+  @doc """
+  Update documents using a function expression.
+  """
+  def update_documents_with_function(%MeilisearchSink{} = sink, index_name, filter, function_expression, context \\ %{}) do
```
acco: 👍

test/sequin/meilisearch_pipeline_test.exs
```diff
@@ -69,6 +69,135 @@ defmodule Sequin.Runtime.MeilisearchPipelineTest do
         record["test"]["id"] == Enum.at(ids, i)
       end)
     end
+
+    @tag :skip
```
acco: Just noting, i assume wip?

davoclavo: yup, I had some changes that are in the works and haven't pushed.
(This comment applies to a couple of other ones, including the Rhai one)

I appreciate the notes tho!

assets/svelte/sinks/meilisearch/MeilisearchSinkForm.svelte
davoclavo: Good idea

test/sequin/meilisearch_pipeline_test.exs
acco: Thanks for pushing up so I could have an early look! 🙏

lib/sequin/runtime/routing/consumers/meilisearch.ex
```diff
@@ -3,30 +3,47 @@ defmodule Sequin.Runtime.Routing.Consumers.Meilisearch do
   use Sequin.Runtime.Routing.RoutedConsumer
 
   @primary_key false
-  @derive {Jason.Encoder, only: [:index_name]}
+  @derive {Jason.Encoder, only: [:index_name, :filter, :function, :context]}
   typed_embedded_schema do
-    field :action, Ecto.Enum, values: [:index, :delete]
+    field :action, Ecto.Enum, values: [:index, :delete, :function]
     field :index_name, :string
+    field :filter, :string
```
davoclavo, line 10: Technically we can omit the filter and it will update all records in the index, however I feel like for the use cases where Sequin will be used it makes sense to have a filter to update a single record (or at most a couple)

lib/sequin/runtime/routing/consumers/meilisearch.ex
```diff
@@ -3,30 +3,47 @@ defmodule Sequin.Runtime.Routing.Consumers.Meilisearch do
...
   end
 
   def route(action, _record, _changes, _metadata) do
     meilisearch_action =
       case action do
-        "insert" -> :index
-        "update" -> :index
-        "delete" -> :delete
-        "read" -> :index
+        "insert" -> "index"
```
davoclavo, line 43: We had inconsistencies with the expected values between here+pipeline batcher, and our sample functions

lib/sequin/sinks/meilisearch/client.ex
davoclavo: Great question, here is an example 400 status code response. I added additional status code checks to handle these errors properly.

```elixir
 %Req.Response{
   status: 400,
   headers: %{
     "content-type" => ["application/json"],
     "date" => ["Mon, 21 Jul 2025 17:07:26 GMT"],
     "transfer-encoding" => ["chunked"],
     "vary" => ["Origin, Access-Control-Request-Method, Access-Control-Request-Headers",
      "accept-encoding"]
   },
   body: %{
     "code" => "invalid_document_filter",
     "link" => "https://docs.meilisearch.com/errors#invalid_document_filter",
     "message" => "Was expecting an operation `=`, `!=`, `>=`, `>`, `<=`, `<`, `IN`, `NOT IN`, `TO`, `EXISTS`, `NOT EXISTS`, `IS NULL`, `IS NOT NULL`, `IS EMPTY`, `IS NOT EMPTY`, `CONTAINS`, `NOT CONTAINS`, `STARTS WITH`, `NOT STARTS WITH`, `_geoRadius`, or `_geoBoundingBox` at `true`.\n1:5 true",
     "type" => "invalid_request"
   },
   trailers: %{},
   private: %{}
 }}
 ```

File: lib/sequin/runtime/meilisearch_pipeline.ex (hunk 2 above)
Comment (davoclavo) [reply lib/sequin/runtime/meilisearch_pipeline.ex]: good point, I just adjusted the pipeline to have a single batcher

File: lib/sequin/runtime/meilisearch_pipeline.ex [hunk 9]
```diff
@@ -23,43 +23,40 @@ defmodule Sequin.Runtime.MeilisearchPipeline do
...
+      |> Broadway.Message.update_data(fn data ->
+        Map.put(data, :routing_info, routing_info)
+      end)
 
     {:ok, message, context}
   end
 
   @impl SinkPipeline
-  def handle_batch(:default, messages, batch_info, context) do
+  def handle_batch(:default, messages, %{batch_key: {"index", index_name}}, context) do
```
acco, line 54: Great!

lib/sequin/runtime/meilisearch_pipeline.ex
```diff
@@ -121,6 +116,53 @@ defmodule Sequin.Runtime.MeilisearchPipeline do
...
+        %{routing_info: routing_info} = message.data
+
+        case Client.update_documents_with_function(
+               sink,
+               index_name,
+               routing_info.filter,
+               routing_info.function,
+               routing_info.context || %{}
+             ) do
+          {:ok} ->
```
acco: Not typical in this codebase, may want to update `Client.update_documents_with_function` to return `:ok`

acco (approved):
All looking good, great work!

lib/sequin/runtime/meilisearch_pipeline.ex
davoclavo: woops, thanks for the catch!

PR #2024, opened 2025-07-18
✨ GCP PubSub batch by messages and bytes

RTLS:
Unfortunately the Broadway batcher will [include the current message in the batch when the batcher responds with `:emit`](https://github.com/dashbitco/broadway/blob/136bea6786ae1526721a98a93ca9d752543c3a7d/lib/broadway/topology/batcher_stage.ex#L159). This makes it hard to enforce a hard byte size ceiling.

I use the `incoming_bytes` as a buffer as a partial workaround -- but this is not perfect.

Considering a PR upstream to move the `event` to `remaining` instead of `acc` -- curious yalls thoughts!

### DIFF
--- a/lib/sequin/runtime/gcp_pubsub_pipeline.ex
+++ b/lib/sequin/runtime/gcp_pubsub_pipeline.ex
@@ -17,12 +17,16 @@ defmodule Sequin.Runtime.GcpPubsubPipeline do
     Map.put(context, :pubsub_client, GcpPubsubSink.pubsub_client(sink))
   end
 
+  @max_bytes Sequin.Size.mb(10)
+
   @impl SinkPipeline
   def batchers_config(consumer) do
     [
       default: [
         concurrency: 400,
-        batch_size: consumer.batch_size,
+        max_demand: consumer.batch_size,
+        # Use 90% of max bytes for buffer
+        batch_size: SinkPipeline.batcher(consumer.batch_size, @max_bytes * 0.9),
         batch_timeout: 1
       ]
     ]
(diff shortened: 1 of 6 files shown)


lib/sequin/runtime/sink_pipeline.ex
```diff
@@ -387,6 +387,30 @@ defmodule Sequin.Runtime.SinkPipeline do
...
+        incoming_bytes = consumer_message.encoded_data_size_bytes
+
+        # This calculation uses the incoming bytes as a buffer (> incoming_bytes)
+        # This is because Broadway will emit the current message in the batch if we return :emit
+        # What we want is to emit the currently accumulated batch and start the next batch with the current message
+        if bytes_remaining - incoming_bytes > incoming_bytes do
+          {:cont, {msgs_remaining - 1, bytes_remaining - incoming_bytes}}
+        else
+          {:emit, {max_messages, max_bytes}}
+        end
```
RTLS, line 408: My workaround is to include the `incoming_bytes` as a buffer. This is strictly correct if all messages have the same size but can lead to batches that violate `max_bytes` if messages have variable byte size.

lib/sequin/runtime/gcp_pubsub_pipeline.ex
```diff
@@ -17,12 +17,15 @@ defmodule Sequin.Runtime.GcpPubsubPipeline do
     Map.put(context, :pubsub_client, GcpPubsubSink.pubsub_client(sink))
   end
 
+  @max_bytes Sequin.Size.mb(10)
```
acco, line 20: I wonder if a littttle wiggle room is good here

lib/sequin/runtime/gcp_pubsub_pipeline.ex
```diff
@@ -38,22 +41,31 @@ defmodule Sequin.Runtime.GcpPubsubPipeline do
     group_id = message.data.group_id
     ordering_key = {topic_id, group_id}
 
-    message = Broadway.Message.put_batch_key(message, ordering_key)
+    encoded_data = build_pubsub_message(context.consumer, message.data)
+    encoded_data_size_bytes = pubsub_message_byte_size(encoded_data)
```
acco, line 46: Cool - looks like we can just build ahead of time because we send the messages all individually anyway (vs one big JSON array)

That makes our calculations pristine!

lib/sequin/runtime/gcp_pubsub_pipeline.ex
```diff
@@ -66,35 +78,41 @@ defmodule Sequin.Runtime.GcpPubsubPipeline do
...
 
-    Sequin.Map.put_if_present(msg, :ordering_key, record.group_id)
+    Sequin.Map.put_if_present(msg, "orderingKey", record.group_id)
   end
 
   defp build_pubsub_message(consumer, %Sequin.Consumers.ConsumerEvent{} = event) do
     msg = %{
-      data: Message.to_external(consumer, event),
-      attributes: %{
+      "data" => Base.encode64(Jason.encode!(Message.to_external(consumer, event))),
```
acco, line 95: Is there an extra encode64 here? Trying to trace where this was before

lib/sequin/runtime/sink_pipeline.ex
acco: That's pretty good, and we can even incorporate some buffer in the global batch limit (i.e. make it 8 or 9MB)

I do think a merge upstream could be valuable as well!

lib/sequin/runtime/gcp_pubsub_pipeline.ex
acco: Never mind, found it! Looks like this is from `lib/sequin/sinks/gcp/pubsub.ex`

test/sequin/gcp_pubsub_pipeline_test.exs
```diff
@@ -0,0 +1,154 @@
...
+        assert length(body["messages"]) == 2
+
+        Req.Test.json(conn, %{})
+      end)
+
+      start_pipeline!(consumer)
+
+      ref = send_test_batch(consumer, [message1, message2])
+      assert_receive {:ack, ^ref, [_message1, _message2], []}, 1_000
+    end
```
acco, line 72: sweet

lib/sequin/runtime/sink_pipeline.ex
acco: Only other consideration is looking at if the message has ever failed - if it has, it ought to be in own batch?

But it seems like that doesn't play nicely with the batcher paradigm here

RTLS: I think that would have to be a different `batcher` or `batch_key` instead of using the `batch_size` function

PR #2025, opened 2025-07-19
change meilisearch request method (document replacement)

fcannizzaro:
Hello! Watching @acco's  video about enrichment I noticed that issue with the meilisearch partial update caused by the use of the `PUT` method, instead of the `POST` one that add/replace a document. 

In the PR I only changed the `PUT` with the `POST` method, since the way it's invoked is the same,  and also updated the documentation / related tests.

I believe an user expects that documents are replaced by default and not merged with their previous versions.


davoclavo:
@fcannizzaro Thanks a lot for your contribution! and also appreciate your patience to get this shipped

PR #2029, opened 2025-07-21
✨ Postgres sink

RTLS:


RTLS:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/sequinstream/sequin/2029?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#2029** <a href="https://app.graphite.dev/github/pr/sequinstream/sequin/2029?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/sequinstream/sequin/2029?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2028** <a href="https://app.graphite.dev/github/pr/sequinstream/sequin/2028?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

lib/sequin/runtime/postgres_pipeline.ex
```diff
@@ -0,0 +1,40 @@
...
+    [default: [concurrency: 10, batch_size: 1, batch_timeout: 1]]
+  end
+
+  @impl SinkPipeline
+  def handle_message(message, context) do
+    {:ok, message, context}
+  end
+
+  @impl SinkPipeline
+  def handle_batch(:default, messages, _batch_info, context) do
```
davoclavo, line 27: Just a reminder to call the routing logic here in the pipeline

lib/sequin/sinks/postgres/implementation.ex
```diff
@@ -0,0 +1,36 @@
...
+    action = record[:action]
+    record = record[:record]
+    changes = record[:changes]
+    metadata = record[:metadata]
+
+    with {:ok, conn} <- ConnectionCache.connection(sink),
+         {:ok, _} <-
+           Postgrex.query(
+             conn,
+             "INSERT INTO #{sink.table_name} (action, record, changes, metadata) VALUES ($1, $2, $3, $4)",
```
davoclavo, line 20: I guess we should figure out a way to escape the table_name to avoid sql injections?

davoclavo: Oh I noticed you added a validation to not include whitespace, good idea

In case a more restrictive validation were useful, here is the official definition of an identifier from the PG docs:

> SQL identifiers and key words must begin with a letter (a-z, but also letters with diacritical marks and non-Latin letters) or an underscore (_). Subsequent characters in an identifier or key word can be letters, underscores, digits (0-9), or dollar signs ($). Note that dollar signs are not allowed in identifiers according to the letter of the SQL standard, so their use might render applications less portable. The SQL standard will not define a key word that contains digits or starts or ends with an underscore, so identifiers of this form are safe against possible conflict with future extensions of the standard.


https://www.postgresql.org/docs/17/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS

danola-sk:
@RTLS any news to share about this? I'm using aws dms currently and it's a huge pain in the ass...

PR #2032, opened 2025-07-24
🛠️ Lazily load message data in Messages tab

### LINKED ISSUE #1912: Lazy‑load large message payloads in the Messages tab
> Defer loading of large payloads from `SlotMessageStore` until the user clicks a message

**Problem**

Opening the **Messages** tab currently copies the full payload of each message from `SlotMessageStore`  into the UI layer. When a single message is abnormally large this duplicate allocation can explode memory usage.

**Solution / Scope**

* Add a metadata‑only read path to `SlotMessageStore`.
* Return only metadata (id, timestamp, sink, size, etc.) to the UI list view.
* Fetch the full payload lazily **only** when a user expands/clicks the row.

davoclavo:
Fixes #1912

### DIFF
--- a/assets/svelte/consumers/ShowMessages.svelte
+++ b/assets/svelte/consumers/ShowMessages.svelte
@@ -215,7 +215,8 @@
         if (reply.error) {
           messageDataError = reply.error;
         } else {
-          messageData = reply.data;
+          selectedMessage = reply.message;
+          messageData = reply.message.data;
         }
       },
     );
--- a/lib/sequin/runtime/slot_message_store_state.ex
+++ b/lib/sequin/runtime/slot_message_store_state.ex
@@ -452,11 +452,30 @@ defmodule Sequin.Runtime.SlotMessageStore.State do
     |> Map.values()
   end
 
+  @doc """
+  Returns messages with metadata only (data field set to nil).
+  This is more memory efficient for listing messages in the UI.
+  """
+  def peek_messages_metadata(%State{} = state, count) when is_integer(count) do
+    state
+    |> sorted_message_stream()
+    |> Enum.take(count)
+    |> Enum.map(&drop_message_data/1)
+  end
+
   def peek_message(%State{} = state, ack_id) do
     case Map.get(state.ack_ids_to_cursor_tuples, ack_id) do
       nil -> {:error, Error.not_found(entity: :message, params: %{ack_id: ack_id})}
       cursor_tuple -> {:ok, Map.get(state.messages, cursor_tuple)}
     end
+  end
+
+  defp drop_message_data(%ConsumerRecord{} = message) do
+    %{message | data: nil}
+  end
+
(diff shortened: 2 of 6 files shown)


lib/sequin/runtime/slot_message_store_state.ex
```diff
@@ -452,13 +452,32 @@ defmodule Sequin.Runtime.SlotMessageStore.State do
...
+  @doc """
+  Returns messages with metadata only (data field set to nil).
+  This is more memory efficient for listing messages in the UI.
+  """
+  def peek_messages_metadata(%State{} = state, count) when is_integer(count) do
+    state
+    |> sorted_message_stream()
+    |> Enum.take(count)
+    |> Enum.map(&strip_message_data/1)
+  end
```
acco: Great - consider `delete_message_metadata` or `drop_`, as `delete`/`drop` aligns with Elixir standard lib (eg `Map.drop`)

assets/svelte/consumers/ShowMessages.svelte
```diff
@@ -215,7 +215,8 @@
         if (reply.error) {
           messageDataError = reply.error;
         } else {
-          messageData = reply.data;
+          selectedMessage = reply.message;
+          messageData = reply.message.data;
```
acco, line 219: nice

test/sequin/slot_message_store_state_test.exs
```diff
@@ -909,6 +909,36 @@ defmodule Sequin.Runtime.SlotMessageStoreStateTest do
     end
   end
 
+  describe "peek_messages_metadata/2" do
+    test "returns messages with data field stripped", %{state: state} do
```
acco: great

test/sequin/slot_message_store_test.exs
```diff
@@ -1187,4 +1187,81 @@ defmodule Sequin.SlotMessageStoreTest do
...
+      [first, second, third] = metadata_messages
+      assert first.commit_lsn == 1
+      assert first.group_id == "test-group-1"
+      assert second.commit_lsn == 2
+      assert second.group_id == "test-group-2"
+      assert third.commit_lsn == 3
+      assert third.group_id == "test-group-3"
+    end
+
+    test "peek_messages returns full messages while peek_messages_metadata returns without data", %{consumer: consumer} do
```
acco: if we have individual tests that assert peek_messages has `data` and `peek_messages_metadata` doesn't have `data`, then this test is probably not necessary

lib/sequin/runtime/slot_message_store_state.ex
davoclavo: @acco just to confirm, you mean renaming the private fn `defp strip_message_data` to `defp drop_message_data` right?  (if so, I just pushed those changes)

or do you mean to find a better name for the public fn `def peek_messages_metadata`??

acco (approved):

PR #2035, opened 2025-07-25
🛠️ Persist state of Pause Updates across tab switches

### LINKED ISSUE #2027: "Pause update" button resets if you leave the Messages tab and come back

davoclavo:
Fixes #2027

### DIFF
--- a/assets/svelte/consumers/ShowSinkHeader.svelte
+++ b/assets/svelte/consumers/ShowSinkHeader.svelte
@@ -35,12 +35,14 @@
   import StopSinkModal from "./StopSinkModal.svelte";
   import { Badge } from "$lib/components/ui/badge";
 
+  type LiveAction = "show" | "backfills" | "messages" | "trace";
   export let consumer;
   export let consumerTitle;
-  export let live_action;
+  export let live_action: LiveAction;
   export let live;
   export let parent;
   export let messages_failing;
+  export let tab_urls;
 
   let showDeleteConfirmDialog = false;
   let showStopModal = false;
@@ -103,32 +105,6 @@
       handleStatusTransition();
     });
   }
-
-  let activeTab: string;
-
-  $: backfillUrl = `${consumer.href}/backfills`;
-
-  $: messageUrl = messages_failing
-    ? `${consumer.href}/messages?showAcked=false`
-    : `${consumer.href}/messages`;
-
-  $: traceUrl = `${consumer.href}/trace`;
-
-  onMount(() => {
-    switch (live_action) {
-      case "backfills":
-        activeTab = "backfills";
-        break;
-      case "messages":
-        activeTab = "messages";
-        break;
-      case "trace":
-        activeTab = "trace";
-        break;
-      default:
-        activeTab = "overview";
-    }
-  });
 
   let sinkDocsSlug: string;
   switch (consumer.sink.type) {
@@ -266,9 +242,9 @@
   <div class="container mx-auto px-4">
     <div class="flex space-x-4">
       <a
-        href={consumer.href}
+        href={tab_urls.overview}
         class={`py-2 px-4 font-medium border-b-2 ${
-          activeTab === "overview"
+          live_action === "show"
             ? "text-black border-black"
             : "text-gray-500 hover:text-gray-700 border-transparent"
         }`}
@@ -278,9 +254,9 @@
         Overview
       </a>
       <a
-        href={backfillUrl}
+        href={tab_urls.backfills}
         class={`py-2 px-4 flex items-center font-medium border-b-2 ${
-          activeTab === "backfills"
+          live_action === "backfills"
             ? "text-black border-black"
             : "text-gray-500 hover:text-gray-700 border-transparent"
         }`}
@@ -295,9 +271,9 @@
         {/if}
       </a>
       <a
-        href={messageUrl}
+        href={tab_urls.messages}
         class={`py-2 px-4 flex items-center font-medium border-b-2 ${
-          activeTab === "messages"
+          live_action === "messages"
             ? "text-black border-black"
             : "text-gray-500 hover:text-gray-700 border-transparent"
         }`}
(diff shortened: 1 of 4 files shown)


lib/sequin_web/live/sink_consumers/show.ex
```diff
@@ -621,14 +626,6 @@ defmodule SequinWeb.SinkConsumersLive.Show do
     {:noreply, assign(socket, trace: %{initial_trace() | paused: false})}
   end
 
-  def handle_event("trace_toggle_show_acked", %{"show_acked" => show_acked}, socket) do
```
davoclavo, line 630: deleting this as it was not being used

assets/svelte/consumers/ShowSinkHeader.svelte
```diff
@@ -106,13 +108,38 @@
...
+  $: backfillUrl = buildUrlWithState(`${consumer.href}/backfills`);
+  $: overviewUrl = buildUrlWithState(consumer.href);
+
+  $: messageUrl = (() => {
+    // Override show_acked to false if messages are failing
+    const effectiveShowAcked = messages_failing ? false : show_acked;
+    return buildUrlWithState(`${consumer.href}/messages`, effectiveShowAcked);
+  })();
+
+  $: traceUrl = buildUrlWithState(`${consumer.href}/trace`);
```
acco: We do most of our URL building elsewhere on the backend instead of the frontend (`handle_params`). This keeps the frontend focused on templating and animations, backend on all business logic.

Possible to do here?

acco (changes requested):
Great idea to use URL params. Can we use them on the backend instead of frontend?

assets/svelte/consumers/ShowSinkHeader.svelte
davoclavo: Good point! let me change this

davoclavo: Done!

lib/sequin_web/live/sink_consumers/show.ex
```diff
@@ -1731,6 +1730,42 @@ defmodule SequinWeb.SinkConsumersLive.Show do
...
+      [] -> ""
+      _ -> "?" <> URI.encode_query(params)
+    end
+  end
+
+  defp push_patch_with_query_params(socket, url_prefix) do
+    %{show_acked: show_acked, paused: paused, consumer: consumer} = socket.assigns
+    query_string = build_query_string(show_acked, paused)
+    push_patch(socket, to: RouteHelpers.consumer_path(consumer, url_prefix <> query_string))
+  end
```
acco: I believe you can use the p sigil to simplify! You can pass query params as a map or kw list:

https://hexdocs.pm/phoenix/routing.html#more-on-verified-routes

Perhaps you can update `RouterHelpers.consumer_path` to accept query params

lib/sequin_web/live/sink_consumers/show.ex
```diff
@@ -1731,6 +1730,42 @@ defmodule SequinWeb.SinkConsumersLive.Show do
...
+
+  defp build_tab_urls(consumer, paused, show_acked, messages_failing) do
```
acco, line 1753: This is fine, though I wonder if there's another way to do this 🤔

lib/sequin_web/live/sink_consumers/show.ex
davoclavo: oh good call! I didn't know about the `~p` sigil, it does make the code a lot simpler

lib/sequin_web/live/sink_consumers/show.ex
davoclavo: good call, I can definitely remove that inner function/closure and keep things simpler, also using the updated `RouteHelpers.consumer_path` with the p sigil

```diff
   defp build_tab_urls(consumer, paused, show_acked, messages_failing) do
-    build_url = fn subpath, override_show_acked ->
-      effective_show_acked = if is_nil(override_show_acked), do: show_acked, else: override_show_acked
-      query_params = build_query_params(effective_show_acked, paused)
-      RouteHelpers.consumer_path(consumer, subpath, query_params)
-    end
+    base_params = build_query_params(show_acked, paused)
+    messages_params = build_query_params(if(messages_failing, do: false, else: show_acked), paused)
 
     %{
-      overview: build_url.("", nil),
-      backfills: build_url.("/backfills", nil),
-      messages: build_url.("/messages", if(messages_failing, do: false)),
-      trace: build_url.("/trace", nil)
+      overview: RouteHelpers.consumer_path(consumer, "", base_params),
+      backfills: RouteHelpers.consumer_path(consumer, "backfills", base_params),
+      messages: RouteHelpers.consumer_path(consumer, "messages", messages_params),
+      trace: RouteHelpers.consumer_path(consumer, "trace", base_params)
     }
   end
```

lib/sequin_web/live/sink_consumers/show.ex
davoclavo: I just noticed that using a very general `~p` sigil causes a warning because of the multiple dynamic segments in the url

```
    warning: no route path for SequinWeb.Router matches "/sinks/#{type}/#{id}/#{subpath}?#{query_params}"
    │
 10 │     ~p"/sinks/#{type}/#{id}/#{subpath}?#{query_params}"
    │       ~
    │
    └─ lib/sequin_web/route_helpers.ex:10: SequinWeb.RouteHelpers.consumer_path/3
```

So I think in order to avoid that warning I will need to break it down into the multiple variants of the `subpath`. Is this OK? or should I backtrack to the `URI.encode_query` approach?

```elixir
  def consumer_path(%SinkConsumer{id: id, type: type}, subpath, query_params) do
    case subpath do
      "" -> ~p"/sinks/#{type}/#{id}?#{query_params}"
      "backfills" -> ~p"/sinks/#{type}/#{id}/backfills?#{query_params}"
      "messages" -> ~p"/sinks/#{type}/#{id}/messages?#{query_params}"
      "messages/" <> ack_id -> ~p"/sinks/#{type}/#{id}/messages/#{ack_id}?#{query_params}"
      "trace" -> ~p"/sinks/#{type}/#{id}/trace?#{query_params}"
      "edit" -> ~p"/sinks/#{type}/#{id}/edit?#{query_params}"
    end
  end
```

acco: Either one sounds good to me! I don't mind the sub-paths, it's nice and explicit

PR #2038, opened 2025-07-25
🛠️ Add retry mechanism for HTTP retryable responses

### LINKED ISSUE #1911: Req should retry requests at least once when we get a retryable error
We use Req as our HTTP lib in many sinks. We should configure it so that it retries at least once when we get a retryable error. Most errors writing to HA destinations will be in this camp, and should have minimal impact on speed throughput/performance

davoclavo:
Use [Req's retry mechanisms](https://hexdocs.pm/req/Req.html#new/1) for [responses/errors that are retryable](https://github.com/wojtekmach/req/blob/main/lib/req/steps.ex#L2175-L2194), such as:

- HTTP status codes 408, 429, 500, 502, 503, 504
- HTTP transport errors: `:timeout`, `:econnrefused`, `:closed`
- We will retry any HTTP verb, as most operations are done with POST/PUT/DELETE, and rely on the idempotency key of messages
- We will retry at most once with a backoff of 500ms -- We could increment the number of retries, and possibly reduce the step number (e.g. 200ms - 3 retries or something)

Fixes #1911

### DIFF
--- a/test/sequin/meilisearch_client_test.exs
+++ b/test/sequin/meilisearch_client_test.exs
@@ -244,4 +244,124 @@ defmodule Sequin.Sinks.Meilisearch.ClientTest do
       assert error.message == "[meilisearch]: Invalid filter expression"
     end
   end
+
+  describe "retry behavior" do
+    test "wait_for_task retries while task is processing" do
+      test_pid = self()
+      call_count = :counters.new(1, [])
+      records = [SinkFactory.meilisearch_record()]
+
+      # First expect the document PUT request
+      Req.Test.expect(Client, fn conn ->
+        assert conn.method == "PUT"
+        assert conn.request_path == "/indexes/test/documents"
+
+        Req.Test.json(conn, %{"taskUid" => 123})
+      end)
+
+      # Then stub the task status checks with multiple retries
+      Req.Test.stub(Client, fn conn ->
+        assert conn.method == "GET"
+        assert conn.request_path == "/tasks/123"
+
+        count = :counters.get(call_count, 1)
+        :counters.add(call_count, 1, 1)
+        send(test_pid, {:task_check, count})
+
+        cond do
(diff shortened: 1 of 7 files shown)


test/sequin/elasticsearch_pipeline_test.exs
```diff
@@ -150,6 +150,54 @@ defmodule Sequin.Runtime.ElasticsearchPipelineTest do
       assert_receive {:ack, _ref, successful, []}, 3000
       assert length(successful) == 3
     end
+
+    @tag capture_log: true
+    test "retries transient errors once before succeeding", %{consumer: consumer} do
```
acco: I don't think these tests are necessary! They fall under the nature of "Does Req do what it says"

test/sequin/meilisearch_client_test.exs
```diff
@@ -130,4 +130,151 @@ defmodule Sequin.Sinks.Meilisearch.ClientTest do
       assert error.message == "[meilisearch]: Index verification failed"
     end
   end
+
+  describe "retry behavior" do
```
acco: As per above, this can be slimmed down or eliminated

acco (changes requested):

test/sequin/elasticsearch_pipeline_test.exs
davoclavo: I agree, however there are some tricky things with Req configurations where I believe having tests that make sure that _our_ config is working as expected is useful.

For example, setting both a `retry_delay:` function  + a `retry:` function causes runtime errors, and that behavior is not documented.

Anyhow, I agree with the general idea of keeping tests to the most important pieces. I will work on trimming this test suite to only keep the essential tests.

acco: Fair!

PR #2039, opened 2025-07-26
add mysql sink

yordis:
- **feat: add mysql support to sink consumer**
- **feat: implement dynamic routing for MySQL sink consumer**


RTLS (commented):
This is incredible work @yordis ! Looks excellent overall and quite close to mergeable.

## UX

- My first pass through was great! Worked exactly as I expected for an insert table using no transform and no routing function into a table with columns `id action record changes metadata` 🔥
- `table_name` is the right parameter and is good to have it available in a routing function
- nit: Layout of mysql connection form is inconsistent with other forms. Each form field should be full width

## Upserts

- Great to give users the option of using `ON DUPLICATE KEY UPDATE`
- This behavior is surprising to me:
    
    ```jsx
    In general, you should try to avoid using an ON DUPLICATE KEY UPDATE clause on tables with multiple unique indexes.
    ```
    

## Casting

- It looks like types are cast based on the value of the record. The supported types are string, boolean, number, and JSONB. Curious how DateTimes work? Bytea columns? What is good about your approach: it is the simplest solution. What might be improved: more explicit type handling
    - Perhaps we can start with what you have but call out the limitations in the docs
    - One route forward might be an additional field in routing that allows type overrides? Or maybe in the transform instead of routing?
- I believe these additions help:

```diff
@@ -176,6 +176,11 @@ defmodule Sequin.Sinks.Mysql.Client do
               value when is_binary(value) -> value
               value when is_number(value) -> value
               value when is_boolean(value) -> value
+              %DateTime{} = dt -> dt
+              %Date{} = date -> date
+              %Time{} = time -> time
+              %NaiveDateTime{} = naive_dt -> naive_dt
+              %Decimal{} = decimal -> decimal
               value -> Jason.encode!(value)
             end
           end)
```

## Error handling

- On test connection failure I received:
    
    ```jsx
    Connection test failed
    [mysql]: MySQL error: %DBConnection.ConnectionError{message: "connection not available and request was dropped from queue after 2000ms. This means requests are coming in and your connection pool cannot serve them fast enough. You can address this by:\n\n  1. Ensuring your database is available and that you can connect to it\n  2. Tracking down slow queries and making sure they are running fast enough\n  3. Increasing the pool_size (although this increases resource consumption)\n  4. Allowing requests to wait longer by increasing :queue_target and :queue_interval\n\nSee DBConnection.start_link/2 for more information\n", severity: :error, reason: :queue_timeout}
    ```
    
    - The issue is obviously not about queueing, requests coming in too fast etc. so converting this error to “test connect failed; check connection details” (or even some trouble shooting steps) would be good.

## Bugs

- compilation error in mysql.ex
- editing a mysql sink erases the password
- routing functions for mysql are not supported in the console
    - ie. can’t create a new routing function for mysql
- when a routing function is added to a mysql sink, the table name should no longer be required on the sink. see `routing_mode` in other sinks for an example.

RTLS (approved):
This looked great last time. If you have addressed the last review and add feature flag to this then merge it 👍

PR #2052, opened 2025-08-04
🛠️ Expose `waiting` snapshot statuses as `ok` status in metrics

davoclavo:


lib/sequin/metrics/entity_health_metrics.ex
```diff
@@ -49,7 +49,7 @@ defmodule Sequin.Metrics.EntityHealthMetrics do
 
     # Map status to the expected values for the gauge
     status_values = [
-      {"ok", snapshot.status == :healthy},
+      {"ok", snapshot.status in [:healthy, :waiting]},
```
RTLS, line 52: nice

RTLS (approved):

Issue #1899, opened 2025-07-01
Clustering configuration is poorly described


zachgersh:
I am trying to self deploy this to a k8s cluster and we've tried almost every type of configuration but our nodes can't seem to be able to communicate with the other node. Could someone post an example k8s deployment or some configuration snippets of how this should look?

https://sequinstream.com/docs/reference/configuration#clustering-configuration

Happy to show more of our deployment config as needed. We are using the sequin-helm-chart.

acco:
Hey @zachgersh –

I just removed the "Clustering" section as it's really not relevant. Sequin doesn't support clustering/distributing across nodes at this time (it was there in a very early version, we just forgot to remove it). But for HA and rolling deploys, it does support [active-passive](https://sequinstream.com/docs/reference/configuration#active-passive-architecture) architecture

We recommend vertically scaling Sequin (it can go very far!) In the near future, we will explore options for horizontal scaling.

PR #2063, opened 2025-08-25
Docs Update - Remove `clustering` reference

seth-acuitymd:
While looking into HA/Production config for Sequin - I noticed this issue - https://github.com/sequinstream/sequin/issues/1899

Based on the reply to the issue, clustering isn't currently supported by Sequin

Further scrolling of docs show a reference to a missing heading for `Clustering configuration` on the `configuration` page is still present - I'm not entirely certain if this bullet point should reference specific config on this page - perhaps the postgres and/or redis vars? or alternatively - this line _could_ be seen as redundant - but I refer to maintainers! 

Thanks team!

### DIFF
--- a/docs/reference/configuration.mdx
+++ b/docs/reference/configuration.mdx
@@ -238,9 +238,6 @@ Requirements for multi-node deployment:
   - The same Postgres database
   - The same Redis instance
   - Identical configuration
-- Nodes must be able to communicate directly with each other
-- Network latency between nodes should be minimal
-- Configuration enviroment variables must be properly set
 
 Sequin uses Redis for distributed locking and leader election.
 
--- a/docs/reference/configuration.mdx
+++ b/docs/reference/configuration.mdx
@@ -240,7 +240,7 @@ Requirements for multi-node deployment:
   - Identical configuration
 - Nodes must be able to communicate directly with each other
 - Network latency between nodes should be minimal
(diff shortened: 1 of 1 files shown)


davoclavo:
@seth-acuitymd Hey Seth, thanks a lot for your contributions! Don't worry about the failing CI check, it was a transient issue. I just ran it again and it passed.

I will let someone else with more context around multi-node deployment approve this PR, but looks good to me!

docs/reference/configuration.mdx
```diff
@@ -240,7 +240,7 @@ Requirements for multi-node deployment:
   - Identical configuration
 - Nodes must be able to communicate directly with each other
 - Network latency between nodes should be minimal
-- Environment variables for [Clustering configuration](#clustering-configuration) must be properly set
+- Configuration enviroment variables must be properly set
```
acco: You can actually delete all these lines!

acco (changes requested):
Hey @seth-acuitymd  - good catch here! This is outdated. We don't do full clustering in active-passive, the two nodes just use Redis as a way to mutex (decide who is active and who is passive).

So, we can delete several lines in this doc

docs/reference/configuration.mdx
seth-acuitymd: Sounds good! Thanks @acco !

PR #2073, opened 2025-09-10
fix: NATS IPv6 connectivity

Gabrola:


Gabrola:
@acco @RTLS any love here? We're having to maintain our own internal fork of Sequin to make sure it works on our ipv6-only stack.

Issue #2078, opened 2025-09-22
Kinesis Sink - AWS Security Token Handling when using ECS Task Role


corleyscotte:
Hey Sequin team,

I discovered what appears to be a bug with the handling of expired security tokens when using the Kinesis sink.
In the logs I am getting the following error:

```bash
[warning] [21:47:50.670] Failed to deliver messages to sink: [aws_kinesis]: Error from AWS: The security token included in the request is expired (status=400) line=506 pid=<0.13227210.0> file=lib/sequin/runtime/sink_pipeline.ex domain=elixir application=sequin account_id=00418827-4530-498c-9aaa-ed754dbb7c05 consumer_id=5ac7a484-9599-4063-b17b-b4bdee53243d
```

The UI also displays the sink as being unhealthy. Initially I thought this may be a permissions issue with the ECS Task Role, however I don't believe that's the case.

Here are the permissions that the role I am using has:
```yml
Policies:
        - PolicyName: KinesisPublisherTaskPolicy
          PolicyDocument:
            Version: '2012-10-17'
            Statement:
              - Effect: Allow
                Action:
                  - kinesis:DescribeStream
                  - kinesis:DescribeStreamSummary
                  - kinesis:ListShards
                  - kinesis:PutRecord
                  - kinesis:PutRecords
                Resource: '*'
        - PolicyName: SequinTaskPolicy
          PolicyDocument:
            Version: '2012-10-17'
            Statement:
              - Effect: Allow
                Action:
                  - ssmmessages:CreateControlChannel
                  - ssmmessages:CreateDataChannel
                  - ssmmessages:OpenControlChannel
                  - ssmmessages:OpenDataChannel
                Resource: '*'
```

The issue can typically be resolved by going through and hitting the refresh icon on the healthcheck for a sink.
I also typically have to go through and hit "redeliver all" to the messages in queue after the healthcheck has confirmed the updated credentials within the container are valid.

Please let me know if you need any additional information, logs, testing, etc.

Thanks!

corleyscotte:
@dosu - Is there an undocumented API endpoint that I can hit to trigger a healthcheck or credential reload? I could use that as a temporary workaround if so.

andriimykhailovsumup:
Hi Sequin team, thanks again for looking into this.

[...]

PR #2104, opened 2026-01-20
refresh AWS credentials if using task role

vagelim:
## what

Adds a refresh to the aws credentials to account for temporary credential expiry when using IAM roles instead of static credentials.

This addresses an issue reported in #2078 

This patch is running in production for me under real loads.

## why

This is a bandaid for a patch that should be applied upstream in the base AWS lib for elixir but I needed to get it working for my case. I am working on a patch to aws-elixir but am not ready to submit.


vagelim:
@acco @RTLS I think this is ready for review 🙏

vagelim:
just wanted to check in on what is needed to get this in @RTLS

vagelim:
We have since moved away from sequin so I will no longer be monitoring this PR.

PR #2107, opened 2026-01-25
add benchmark pipeline

acco:


config/bench.exs
```diff
@@ -0,0 +1,31 @@
+import Config
+
+config :sequin, Oban, testing: :manual, prefix: "sequin_config"
```
yordis, line 3:
```suggestion
config :sequin, Sequin.Oban, testing: :manual, prefix: "sequin_config"
```

🥂 (I created `use Oban` for this reason btw, so no implicit globals)

PR #2114, opened 2026-02-18
Add libcluster_postgres for automatic node discovery

RTLS:
Use PostgreSQL LISTEN/NOTIFY via libcluster_postgres for cluster
formation, removing the need for DNS-based discovery. This enables
distributed features (syn, global, MutexedSupervisor) across nodes
using the existing Postgres database.

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>


lib/sequin/application.ex
```diff
@@ -75,6 +75,7 @@ defmodule Sequin.Application do
     Sequin.Redis.connect_cluster()
 
     [
+      {Cluster.Supervisor, [Application.get_env(:libcluster, :topologies, []), [name: Sequin.ClusterSupervisor]]},
```
yordis, line 78:
```suggestion
      Sequin.ClusterSupervisor,
```

then `Sequin.ClusterSupervisor.child_spec/1` ask acco

PR #2131, opened 2026-03-26
fix: MutexOwner retries on transient Redis errors instead of crashing

### LINKED ISSUE #2072: Redis lost connection does not seem to get re-created
Hello, when running sequin in k8s, it seems every time my redis restarts, sequin gets blocked I and I get these errors : 
```
Last message: {:phoenix, :async_result, {:start, {#Reference<0.0.3396227.629622195.1249705985.119942>, nil, :consumers_task, {:exit, {{:badmatch, {:error, %Sequin.Error.ServiceError{code: "no_connection", message: "[redis]: No connection to Redis", service: :redis, details: nil}}}, [{SequinWeb.SinkConsumersLive.Index, :"-load_consumer_metrics/1-fun-0-", 1, [file: ~c"lib/sequin_web/live/sink_consumers/index.ex", line: 168]}, {Enum, :"-map/2-lists^map/1-1-", 2, [file: ~c"lib/enum.ex", line: 1714]}, {Map, :new_from_enum, 2, [file: ~c"lib/map.ex", line: 262]}, {SequinWeb.SinkConsumersLive.Index, :load_consumers, 3, [file: ~c"lib/sequin_web/live/sink_consumers/index.ex", line: 147]}, {Phoenix.LiveView.Async, :do_async, 5, [file: ~c"lib/phoenix_live_view/async.ex", line: 220]}, {Task.Supervised, :invoke_mfa, 2, [file: ~c"lib/task/supervised.ex", line: 101]}]}}}}} line=2646 pid=<0.26533.0> file=gen_server.erl domain=otp
[error] [08:25:01.621] Redis error while trying to acquire mutex: {:error, %Sequin.Error.ServiceError{code: "no_connection", message: "[redis]: No connection to Redis", service: :redis, details: nil}} line=25 pid=<0.20162.0> file=lib/sequin/mutex.ex domain=elixir application=sequin mutex_key=sequin:mutexed_supervisor:prod:Elixir.Sequin.Runtime.MutexedSupervisor
```

It gets unblocked when I restart the sequin pod. I'm using sequin v0.13.14

isaacrowntree:
## Summary

Fixes #2072 — When Redis (or a Redis-compatible store like Dragonfly/KeyDB) restarts, Sequin enters an unrecoverable failure state requiring a full restart. This has bitten us (and others per #2072) multiple times in production.

**Root cause:** `MutexOwner` in `:has_mutex` state calls `acquire_mutex` which returns `:error` when Redis is unreachable. The handler immediately returns `{:shutdown, :err_keeping_mutex}` — an invalid GenStateMachine return that crashes the process with `{:bad_return_from_state_function, ...}`. Due to `MutexedSupervisor`'s `:one_for_all` strategy, this cascades and takes down the entire `Runtime.Supervisor` including all consumers.

**Changes:**

- **`MutexOwner`**: Instead of crashing on Redis errors, retry indefinitely with exponential backoff (capped at 1 hour). When Redis comes back, the error counter resets and normal operation resumes. Also fixes the GenStateMachine return value (`{:stop, {:shutdown, reason}}` instead of invalid `{:shutdown, reason}`).
- **`SinkConsumersLive.Index`**: Handle `{:error, _}` from metrics calls instead of crashing with `MatchError`.
- **`SinkConsumersLive.Show`**: Same defensive handling for all three metrics calls.
- **`HttpEndpointsLive.Show`**: Same fix for `get_http_endpoint_throughput/1`.

## Context

We self-host Sequin on Railway with Dragonfly (Redis-compatible) as the backing store. Railway periodically auto-updates Dragonfly, which causes a brief restart. Every time this happens, Sequin fails to self-heal and requires a manual restart — we've hit this 3 times now. The `:await_mutex` state already retries correctly on Redis errors; this PR brings the same resilience to the `:has_mutex` state, but with exponential backoff so it doesn't spam during extended outages.

## Test plan

- [x] Added `MutexOwnerTest` with unit tests for State struct and integration tests (tagged `:integration`) that use `iptables REJECT` to simulate Redis going down and coming back
- [x] Integration tests verify: process survives Redis outage, never crashes, recovers when Redis returns
- [x] All existing tests pass (`mix test`)
- [x] `mix format --check-formatted` passes
- [ ] Verify in staging by restarting Redis while Sequin is running

Note: Integration tests require `NET_ADMIN` capability (for iptables) and are tagged `:integration` so they can be excluded from normal test runs: `mix test --exclude integration`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/lib/sequin/mutex_owner.ex
+++ b/lib/sequin/mutex_owner.ex
@@ -5,8 +5,9 @@ defmodule Sequin.MutexOwner do
   If it ever loses the mutex (unexpected - it should be touching the mutex before it expires), it crashes.
 
   If Redis becomes temporarily unreachable while holding the mutex, the MutexOwner will retry
-  with backoff rather than immediately crashing. This handles transient Redis disconnections
-  (e.g. Redis/Dragonfly/KeyDB restarts) without cascading failures through MutexedSupervisor.
+  with exponential backoff indefinitely (capped at 1 hour) rather than crashing. This handles
+  Redis/Dragonfly/KeyDB restarts without cascading failures through MutexedSupervisor.
+  When Redis comes back, the MutexOwner re-acquires the mutex and resumes normal operation.
   """
   use GenStateMachine
 
@@ -14,7 +15,8 @@ defmodule Sequin.MutexOwner do
 
   require Logger
 
-  @max_redis_errors 5
+  # Retry backoff caps at 1 hour
+  @max_retry_interval to_timeout(hour: 1)
 
   defmodule State do
     @moduledoc """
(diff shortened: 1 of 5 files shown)


isaacrowntree:
Since upstream appears unmaintained (no maintainer activity since Feb 2026), this fix is now merged and released in the maintained fork at https://github.com/triptechtravel/sequin — release [v0.14.6-tt2](https://github.com/triptechtravel/sequin/releases/tag/v0.14.6-tt2), image `ghcr.io/triptechtravel/sequin:v0.14.6-tt2`. The fork continues from upstream's final v0.14.6 with production-driven maintenance (security/crash/bug fixes).

PR #2151, opened 2026-07-30
fix(yaml_loader): refresh stale table cache on miss during config apply

### LINKED ISSUE #2150: YAML config apply resolves change_retentions against a stale table cache, crash-looping self-hosted deployments when tables are new
### Summary

When applying declarative YAML config (`Sequin.YamlLoader.apply_from_stdin!` via `start_commands.sh`), `change_retentions` source/destination tables are resolved against the **cached** table list stored on the `postgres_databases` record — not against the live database. The cache is only refreshed on `create_db` and by a 6-hourly cron (`EnqueueDatabaseUpdateWorker`). If a table was created after the last refresh, config apply fails even though the table exists and the connection role can see it:

```
** (RuntimeError) Failed to apply config: %Sequin.Error.BadRequestError{message: "Error setting up change retention '...': Not found: No `table` found matching params: `name=..., schema=public`", code: nil}
```

Because config apply runs at container start (before `start_application`), the failure crash-loops the container — which in turn largely prevents the 6-hourly refresh from ever running, so the deployment cannot self-heal.

### Environment

- Sequin `v0.14.1`, self-hosted, headless (`SERVER_CHECK` disabled — no web console available)
- Config delivered via `CONFIG_FILE_PATH`, applied on every container start
- Source database on Cloud SQL for PostgreSQL 16

### Steps to reproduce

1. Run self-hosted Sequin with a YAML config including a database and `change_retentions`.
2. After Sequin has been running for a while (so `tables_refreshed_at` is in the past), create a new table in the source database.
3. Add a `change_retentions` entry for the new table and restart the container so the config re-applies.
4. Config apply fails with `No table found matching params`, the container exits, and it keeps crash-looping until the cache happens to be refreshed.

### Where this happens in the code

- `lib/sequin/yaml_loader.ex` — `parse_wal_pipeline_attrs/2` calls `fetch_table(source_database.tables, ...)`, which searches the cached list only.
- `lib/sequin/databases/databases.ex` — `update_db/2` (the path taken for an existing database during config apply) never calls `update_tables/1`.
- `config/config.exs` — the only periodic refresh is `{"0 */6 * * *", Sequin.Databases.EnqueueDatabaseUpdateWorker}`.

### Workarounds we tried

- Granting privileges / verifying the table is visible to the replication role: no effect (the lookup never touches the live DB).
- Hand-inserting an `Oban` job for `Sequin.Databases.DatabaseUpdateWorker` into `oban_jobs`: the job was never picked up inside the short crash-loop windows (manual inserts don't fire the insert notification, and flipping the row to `scheduled` for the stager didn't get it dispatched in time either).
- What finally works is either manually writing the table entry into the `postgres_databases.tables` jsonb, or waiting for a boot to happen to straddle the 6-hour cron tick — neither is great operationally.

### Request

Any of the following would solve this class of problem:

1. **Refresh on miss**: when `fetch_table` misses during YAML config apply, call `Databases.update_tables/1` once and retry the lookup before failing. This keeps the fast path cached but makes declarative config deterministic.
2. **Refresh on apply**: unconditionally refresh tables for each upserted database during `apply_from_yml` (config apply is rare and already talks to the database).
3. **An operator escape hatch**: a documented release task, e.g. `bin/sequin eval "Sequin.Release.refresh_tables"`, and/or a CLI command (`sequin databases refresh-tables`) so headless deployments can force a refresh without touching internal tables.

Option 1 seems the most aligned with the existing design. Happy to contribute a PR if you'd accept one — thanks for the great tool!

ddavignon-hadrian:
## Problem

Fixes #2150.

When applying declarative YAML config, `change_retentions` resolve their source/destination tables against the `tables` list cached on the `postgres_databases` record. That cache is only refreshed when a database is **created** and by the 6-hourly `EnqueueDatabaseUpdateWorker` cron — the config-apply update path (`update_db/2`) never refreshes it. So if a table is created after the last refresh, `apply_config` fails with:

```
No `table` found matching params: `name=..., schema=public`
```

Because config apply runs at container start (before `start_application`), this fails the boot and crash-loops the container, which in turn prevents the 6-hourly cron from ever running — so the deployment can't self-heal. (Full analysis in #2150.)

## Fix

Refresh the referenced database's tables from the live database on a cache miss while parsing `change_retention` attrs, then retry the lookup (option 1 from the issue). A new `ensure_table_cached/3` helper:

- returns the database unchanged on a cache **hit** (existing in-memory fast path — no extra round-trip);
- on a **miss**, calls `Databases.update_tables/1` once to refresh from the live DB, preserving the `replication_slot` preload;
- still returns `not_found` if the table is genuinely absent after refreshing (no behavior change for real misses).

A round-trip is only added when a referenced table is missing from the cache, so applies that don't reference new tables are unchanged.

## Test

Adds a `change_retentions` test that creates a database, clears its cached `tables` to simulate a stale cache, then applies a config whose `change_retention` references those tables and asserts the apply succeeds and the pipeline is created.

### DIFF
--- a/lib/sequin/yaml_loader.ex
+++ b/lib/sequin/yaml_loader.ex
@@ -727,6 +727,8 @@ defmodule Sequin.YamlLoader do
        ) do
     with {:ok, source_database} <- fetch_database(databases, source_db_name, :source_database),
          {:ok, destination_database} <- fetch_database(databases, dest_db_name, :destination_database),
+         {:ok, destination_database} <- ensure_table_cached(destination_database, dest_schema, dest_table),
+         {:ok, source_database} <- ensure_table_cached(source_database, source_schema, source_table),
          {:ok, destination_table} <-
            fetch_table(destination_database.tables, dest_schema, dest_table, :destination_table),
          {:ok, source_table_struct} <-
@@ -791,6 +793,22 @@ defmodule Sequin.YamlLoader do
     case Enum.find(tables, &(&1.schema == schema && &1.name == table_name)) do
       nil -> {:error, Error.not_found(entity: :table, params: %{schema: schema, name: table_name})}
       table -> {:ok, table}
+    end
+  end
+
+  # The cached `tables` can be stale during config apply (only refreshed on create
+  # and by a 6-hourly cron), so on a miss refresh once from the live database.
+  defp ensure_table_cached(%PostgresDatabase{} = database, schema, table_name) do
+    if Enum.any?(database.tables, &(&1.schema == schema && &1.name == table_name)) do
+      {:ok, database}
+    else
+      case Databases.update_tables(database) do
+        {:ok, refreshed} ->
+          {:ok, %{database | tables: refreshed.tables, tables_refreshed_at: refreshed.tables_refreshed_at}}
+
+        {:error, error} ->
+          {:error, error}
+      end
     end
   end
 
--- a/test/sequin/yaml_loader_test.exs
+++ b/test/sequin/yaml_loader_test.exs
@@ -188,6 +188,53 @@ defmodule Sequin.YamlLoaderTest do
     end
   end
 
+  describe "change_retentions" do
+    test "refreshes a stale table cache on miss during config apply" do
+      # First apply creates the database; create_db refreshes its table cache.
+      assert :ok = YamlLoader.apply_from_yml!(playground_yml())
+
+      [%PostgresDatabase{} = db] = Repo.all(PostgresDatabase)
+
+      # Simulate a stale cache (as if the tables were created after the last
+      # refresh): clear the cached table list on the existing database record.
+      {:ok, _} = db |> PostgresDatabase.changeset(%{tables: []}) |> Repo.update()
+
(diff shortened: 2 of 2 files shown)


ddavignon-hadrian:
Quick heads up on the failing `signoff` check: it died at the Configure PostgreSQL step with `psql: command not found` (exit 127) on the `warp-ubuntu-latest-arm64-32x` runner. That's a setup step, so nothing after it ran and the Elixir compile and tests were skipped entirely. I don't think it's related to this PR, which only changes `lib/sequin/yaml_loader.ex` and its test.

Looks like the runner is missing the postgres client. Could a maintainer re-run it once that's sorted?

In the meantime I ran it locally on Elixir 1.19.4 / OTP 28.3 against Postgres 17 with pgvector and logical replication. The project compiles clean and the new test passes. I also checked that the test fails without the fix, with the same "No table found matching params: name=sequin_events, schema=public" error from the linked issue, so it reproduces the bug and the change resolves it.

Happy to change anything on my side if it's needed.

